Issue #12, opened 2019-11-25
HAL: interrupts abstraction


hawkw:
from #2:
>  + interrupts
>.   - some kinda interface for cross-platform interrupt handlers (presumably not the actual ISRs which are probably pretty platform specific, but the kernel logic they call into)
>.   - enable/disable interrupts

+ seems like it would be good to take cross-platform interrupt handlers and generate platform ISRs (macros?)
+ would probably need an init iface (configure cpu interrupt vectors)

hawkw:
how much interrupt handling should we try to abstract over, and how much is so platform-specific it should just be hidden behind a big "do platform-specific stuff" fn? cc @iximeow

hawkw:
kinda sorta done, or at least started

hawkw:
bones are done as of #15 & the kernel actually defines it's own handlers as of #55. will need additional work in the long run.

PR #15, opened 2019-11-25
feat(hal-core): rough sketch of interrupt APIs 

hawkw:
Depends on #10.

This is a very rough initial sketch of a potential design. See also #12.

### DIFF
--- a/hal-core/src/interrupt.rs
+++ b/hal-core/src/interrupt.rs
@@ -1,0 +1,125 @@
+/// An interrupt controller for a platform.
+pub trait Control {
+    type Error;
+    /// The type of this platform's interrupt vector.
+    ///
+    /// TODO(eliza): This is *probably* always a `u8`, do we really need to have
+    /// this?
+    type Vector;
+
+    /// Disable interrupts.
+    unsafe fn disable_irq(&mut self);
+
+    /// Enable interrupts.
+    unsafe fn enable_irq(&mut self);
+
+    /// Returns `true` if interrupts are enabled.
+    fn is_enabled(&self) -> bool;
+
+    /// Register an interrupt handler to service the provided interrupt.
+    ///
+    /// The interrupt controller is assumed to generate an actual ISR that calls
+    /// the handler; that ISR will construct an interrupt context for the
+    /// handler. The ISR is responsible for entering a critical section while
+    /// the handler is active.
+    fn register_handler<I>(
+        &mut self,
+        irq: &I,
+        handler: I::Handler,
+    ) -> Result<(), RegistrationError>
+    where
+        I: Interrupt<Ctrl = Self>;
+
+    /// Enter a critical section, returning a guard.
+    fn enter_critical(&mut self) -> CriticalGuard<'_, Self> {
+        unsafe {
+            self.disable_irq();
+        }
+        CriticalGuard { ctrl: self }
+    }
+}
+
+/// An interrupt.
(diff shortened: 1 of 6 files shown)


hal-core/src/interrupt.rs
```diff
@@ -0,0 +1,125 @@
...
+pub trait Control {
+    type Error;
+    /// The type of this platform's interrupt vector.
+    ///
+    /// TODO(eliza): This is *probably* always a `u8`, do we really need to have
+    /// this?
+    type Vector;
+
+    /// Disable interrupts.
+    unsafe fn disable_irq(&mut self);
```
iximeow: why is this and `enable_irq` `unsafe`? because irqs (and masking) are global mutable state? i'm perfectly fine with this for my own terrible reasoning i'm just curious

hal-core/src/interrupt.rs
```diff
@@ -0,0 +1,125 @@
+/// An interrupt controller for a platform.
+pub trait Control {
+    type Error;
+    /// The type of this platform's interrupt vector.
+    ///
+    /// TODO(eliza): This is *probably* always a `u8`, do we really need to have
```
iximeow: i like having this open because while we could probably linearize like, two apics into one u8 "address space" of interrupts, it might be nicer to model it as a pair.

hal-core/src/interrupt.rs
```diff
@@ -0,0 +1,125 @@
...
+    fn register_handler<I>(
+        &mut self,
+        irq: &I,
+        handler: I::Handler,
+    ) -> Result<(), RegistrationError>
+    where
+        I: Interrupt<Ctrl = Self>;
+
+    /// Enter a critical section, returning a guard.
+    fn enter_critical(&mut self) -> CriticalGuard<'_, Self> {
```
iximeow: yay this is nice

hal-core/src/interrupt.rs
hawkw: IIRC if you `sti` twice in a row it does a general protection fault or something. maybe i made that up though?

hawkw: upon further consideration it is probably the platform impl’s job to “not do that”; these should probably be safe

hawkw: i also considered having geese take `&mut self` even though they are probably always going to be a single instruction that doesn’t actually _mutate_ the receiver...but maybe the platform impl wants to do bookkeeping as well? idk

hal-core/src/interrupt.rs
hawkw: it could probably also just be called `critical`, but i was thinking about having a version taking a closure. there’s not really a compelling reason to have both though, and the drop guard is a more flexible api

hal-core/src/interrupt.rs
hawkw: I forgot about apic tbh. this should probably not be hardcoded to u8!

hal-core/src/interrupt.rs
iximeow: i can definitely imagine impls that would want to do bookkeeping, yes

hawkw: there should probably also be a `fn mask(Interrupt<Ctrl=Self>)` and a corresponding `unmask`; there could be a drop guard API implemented on top of those as well

hawkw:
I'm gonna merge this & add masking later.

PR #16, opened 2019-11-27
feat(util): add util crate 

hawkw:
This branch adds the beginnings of a `mycelium-util` crate, which is
sort of like a mini `no-std` standard library. In particular, this came
out of wanting an `Error` trait to use in the hal experiments I'm
working on, and not being able to use the `std::error::Error` trait. I
thought putting our error trait in a separate crate from `hal-core` was
a good idea, so that we could use it elsewhere in the kernel.

I also got bored and wrote a quick spinlock. The spinlock is tested
using `loom`, and there's a `loom` feature flag to turn on so that the
spinlock can also participate in other crates' loom tests.


util/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+///
+/// The [`source`] method is generally used when errors cross "abstraction
+/// boundaries". If one module must report an error that is caused by an error
+/// from a lower-level module, it can allow access to that error via the
+/// [`source`] method. This makes it possible for the high-level module to
+/// provide its own errors while also revealing some of the implementation for
+/// debugging via [`source`] chains.
+///
+/// [`source`]: trait.Error.html#method.source
+pub trait Error: fmt::Debug + fmt::Display {
```
mystor: Given that we don't need to perfectly match `std`, perhaps we should include a `'static` bound on `Error`, and avoid the need for such a bound on the other methods. IIRC `std` only doesn't have that bound for backwards-compatibility reasons.

util/src/error.rs
```diff
@@ -0,0 +1,165 @@
...
+        let current = self.current;
+        self.current = self.current.and_then(Error::source);
+        current
+    }
+}
+
+mod private {
+    // This is a hack to prevent `type_id` from being overridden by `Error`
+    // implementations, since that can enable unsound downcasting.
+    #[derive(Debug)]
```
mystor: Why `derive(Debug)` here?

util/src/sync/mutex.rs
```diff
@@ -0,0 +1,126 @@
+use super::atomic::{spin_loop_hint, AtomicBool, Ordering};
+use crate::cell::CausalCell;
+use core::{fmt, ops};
+
+/// A simple spinlock ensuring mutual exclusion.
+#[derive(Debug)]
+pub struct Mutex<T> {
```
mystor: I wonder whether it'd be worthwhile to call this `Spinlock<T>` instead of `Mutex<T>`, to make it clear that it has a spinlock's performance characteristics. It's probably already pretty clear, given that we're in a kernel & don't have a scheduler.

util/src/error.rs
hawkw: Hmm, not a bad idea! i was thinking about experimenting with some other APIs that std doesn’t have here, as well (error categories/severities?)

util/src/error.rs
hawkw: “idk, std does it”

util/src/sync/mutex.rs
hawkw: That’s not a bad call! I was thinking about writing a read-write lock too, though

util/src/error.rs
mystor: Stuff like that would probably be feasible to add with non-defaulted impls in the future, so I don't think we need to worry about getting that right first try. Probably worth having some sort of error classification like that, though.

util/src/sync/mutex.rs
mystor: We could have `SpinLock<T>` (`SpinMutex<T>`?) and `SpinRwlock<T>` if needed. I don't have super strong feelings about this, though.

hawkw: They could also go in a `spin` module, though I don't imagine we'd ever have any _other_ kind of mutex. I don't feel passionately about this either.

PR #21, opened 2019-11-27
feat(hal-core): add page types

hawkw:


hal-core/src/mem/page.rs
```diff
@@ -0,0 +1,204 @@
...
+    type Output = Self;
+    fn sub(self, rhs: usize) -> Self {
+        Page {
+            base: self.base - (S::SIZE * rhs),
+            _size: PhantomData,
+        }
+    }
+}
+
+impl<A: Address, S: Size> AsRef<[u8]> for Page<A, S> {
```
iximeow: i was going to ask if there should be an `AsMut` too, but if we're only reading pages to like, serialize them or something, and we write to pages by "writing to pointers" and letting the mmu deal with it, we don't need `AsMut` huh?

hawkw: upon further consideration i think this should probably be a method rather than an AsRef impl, so that it can be unsafe to call...you would need to ensure nothing is concurrently writing to the page.

iximeow: haha, supporting concurrency.... (this is a very good point)

PR #27, opened 2019-12-12
feat(x86_64): a kernel that boots

### LINKED ISSUE #26: a kernel that boots on x64

hawkw:
closes #26 

this branch adds everything we need to boot the kernel on x86_64, plus
an attempt at organizing the main kernel crate and modules. it adds
stubs of a kernel main function, which currently just prints the version
and architecture. additionally, i added a quick VGA driver
implementation for x86, so we can print stuff.

the x86_64 kernel uses the [`bootloader`] crate by @phil-opp, which
implements a simple bootloader in rust for x86. i thought about using
GRUB, but this is simpler and lets us implement the whole build as 
`cargo` commands, which is nice — it should be quite simple to set
up a dev environment. i'm not sure if this bootloader works for 32-bit
x86 targets or if it's x64-specific, so we may need to figure something
else out to target 32-bit platforms.

sorry if any of the code or docs added in this branch doesn't make sense,
i have the flu and am kinda delirious.

[`bootloader`]: https://github.com/rust-osdev/bootloader

Screen Shot 2019-12-12 at 3 27 42 PM

### DIFF
--- a/hal-core/src/boot.rs
+++ b/hal-core/src/boot.rs
@@ -4,9 +4,13 @@ use core::iter::Iterator;
 pub trait BootInfo {
     type Arch: Architecture;
     type MemoryMap: Iterator<Item = mem::Region<Self::Arch>>;
+    type Writer: core::fmt::Write;
 
     /// Returns the boot info's memory map.
     fn memory_map(&self) -> Self::MemoryMap;
+
+    /// Returns a writer for printing early kernel diagnostics
+    fn writer(&self) -> Self::Writer;
 
     // TODO(eliza): figure out a non-bad way to represent boot command lines (is
     // it reasonable to convert them to rust strs when we barely have an operating
--- a/hal-x86_64/src/lib.rs
+++ b/hal-x86_64/src/lib.rs
@@ -1,0 +1,109 @@
+//! Implementation of the Mycelium HAL for 64-bit x86 platforms.
+#![no_std]
+use core::{fmt, ops};
+use hal_core::{Address, Architecture};
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct PAddr(u64);
+
+#[derive(Debug)]
+pub struct X64;
+
+impl Architecture for X64 {
+    type PAddr = PAddr;
+    const NAME: &'static str = "x86_64";
+}
+
+impl fmt::Debug for PAddr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("PAddr")
+            .field(&format_args!("{:#x}", self.0))
+            .finish()
+    }
+}
+
+impl ops::Add<usize> for PAddr {
+    type Output = Self;
+    fn add(self, rhs: usize) -> Self {
+        PAddr(self.0 + rhs as u64)
+    }
+}
+
+impl ops::Add for PAddr {
+    type Output = Self;
+    fn add(self, rhs: Self) -> Self {
+        PAddr(self.0 + rhs.0)
+    }
+}
+
+impl ops::AddAssign for PAddr {
+    fn add_assign(&mut self, rhs: Self) {
+        self.0 += rhs.0;
+    }
+}
+
+impl ops::AddAssign<usize> for PAddr {
+    fn add_assign(&mut self, rhs: usize) {
+        self.0 += rhs as u64;
+    }
+}
+
+impl ops::Sub<usize> for PAddr {
+    type Output = Self;
+    fn sub(self, rhs: usize) -> Self {
+        PAddr(self.0 - rhs as u64)
+    }
+}
+
+impl ops::Sub for PAddr {
+    type Output = Self;
+    fn sub(self, rhs: Self) -> Self {
+        PAddr(self.0 - rhs.0)
+    }
+}
+
+impl ops::SubAssign for PAddr {
+    fn sub_assign(&mut self, rhs: Self) {
+        self.0 -= rhs.0;
+    }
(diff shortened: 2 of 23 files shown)


phil-opp:
> i'm not sure if this bootloader works for 32-bit
> x86 targets or if it's x64-specific, so we may need to figure something
> else out to target 32-bit platforms.

The `bootloader` currently only support legacy BIOS booting on 64-bit, but we plan to add support for UEFI, the multiboot standard (GRUB integration), and 32-bit x86 eventually.

util/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mycelium-util"
-version = "0.1.0"
+version = "0.0.1"
```
mystor: Why the version drop? Doesn't really matter, but it seems reasonable to leave it at `0.1.0`.

hal-core/src/boot.rs
```diff
@@ -4,10 +4,14 @@ use core::iter::Iterator;
 pub trait BootInfo {
     type Arch: Architecture;
     type MemoryMap: Iterator<Item = mem::Region<Self::Arch>>;
+    type Writer: core::fmt::Write;
```
mystor: We might want to introduce our own wrapper `Write` trait, like `std::io::Write`, which handles proper error propagation. Perhaps in the util library?
It re-wraps formatters to bubble errors out: https://github.com/rust-lang/rust/blob/3964a55ba59bb6c3d4badcbddc49f6929ef76862/src/libstd/io/mod.rs#L1404-L1437

util/src/sync/spin.rs
```diff
@@ -14,13 +17,22 @@ pub struct MutexGuard<'a, T> {
 }
 
 impl<T> Mutex<T> {
+    #[cfg(any(test, feature = "loom"))]
     pub fn new(data: T) -> Self {
```
mystor: I take it `loom` doesn't have a `const fn new` method? It'd be nice to keep the constness similar.

hal-core/src/boot.rs
hawkw: Yeah, I think this is probably a good idea — I want to do this in a follow-up

util/src/sync/spin.rs
hawkw: > I take it `loom` doesn't have a `const fn new` method?

That's correct. Constructing a `loom` atomic or `CausalCell` inside of a test requires accessing the "execution" (test-run-specific data).

We can probably get rid of the `const fn` new actually, since the VGA buffer is in a `lazy_static`; I added this because I was hoping to use a regular static (but you can't do pointer casts in static initializers).

hal-core/src/boot.rs
hawkw: Opened  #29

hawkw:
> > i'm not sure if this bootloader works for 32-bit
> > x86 targets or if it's x64-specific, so we may need to figure something
> > else out to target 32-bit platforms.
> 
> The `bootloader` currently only support legacy BIOS booting on 64-bit, but we plan to add support for UEFI, the multiboot standard (GRUB integration), and 32-bit x86 eventually.

Cool, when we get to the point where we need UEFI or multiboot, or want to support 32-bit targets, I think we might be interested in helping out with that!

util/Cargo.toml
hawkw: The kernel prints the version on boot now (taken from the main package cargo version) and i thought it seemed pretty fake to print 0.1. So, I did a find and replace for all the version numbers. I can change it back if you prefer.

hawkw:
@mystor btw, I've addressed all your reviews!

hal-x86_64/src/lib.rs
```diff
@@ -0,0 +1,111 @@
+//! Implementation of the Mycelium HAL for 64-bit x86 platforms.
+#![cfg_attr(not(test), no_std)]
+use core::{fmt, ops};
+use hal_core::{Address, Architecture};
+
+pub mod vga;
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct PAddr(u64);
```
iximeow: `#[repr(transparent)]`?

hal-x86_64/src/vga.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    fn write(&mut self, s: &str) {
+        assert!(s.is_ascii(), "VGA buffer does not support Unicode");
+        for ch in s.bytes() {
+            self.write_char(ch);
+        }
+    }
+
+    fn write_char(&mut self, ch: u8) {
+        // TODO(eliza): assert char is ascii?
```
iximeow: the thing we're writing to (i believe?) speaks [CP437](https://en.wikipedia.org/wiki/Code_page_437), so any u8 is a valid character to write

hawkw: yeah, i guess if we're making assertions about ascii-ness, it should only be when trying to write a _string slice_ to the vga buffer...i was thinking more "avoid accidentally printing unicode"

hawkw: OH LOL we already do that when printing string slices, nvm.

iximeow: yeah that makes sense. i think checking that in `write(self, &str)` would be better so we can style on the vga buffer later (i almost want to volunteer a tui but i won't say it)

iximeow: of course nothing forbids explicit style functions here to do that

hawkw: > i think checking that in `write(self, &str)` would be better

yeah we already do that :)

PR #33, opened 2019-12-13
feat(x86_64): read memory map from bootloader

hawkw:
depends on #27

the x64 kernel now reads the memory map from the bootloader and converts
it into our HAL types. i also implemented some unimplemented functions
in order to make this work.

Screen Shot 2019-12-13 at 11 07 54 AM

### DIFF
--- a/hal-core/src/mem/mod.rs
+++ b/hal-core/src/mem/mod.rs
@@ -1,9 +1,9 @@
 use crate::Architecture;
-
+use core::fmt;
 pub mod page;
 
 /// A cross-platform representation of a memory region.
-#[derive(Debug, Clone)]
+#[derive(Clone)]
 pub struct Region<A: Architecture> {
     base: A::PAddr,
     // TODO(eliza): should regions be stored as (start -> end) or as
@@ -12,7 +12,7 @@ pub struct Region<A: Architecture> {
     kind: RegionKind,
 }
 
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone, Eq, PartialEq)]
 #[repr(C)]
 pub struct RegionKind(KindInner);
 
@@ -47,6 +47,10 @@ enum KindInner {
 }
 
 impl<A: Architecture> Region<A> {
+    pub fn new(base: A::PAddr, size: usize, kind: RegionKind) -> Self {
+        Self { base, size, kind }
+    }
+
     /// Returns the base address of the memory region
     pub fn base_addr(&self) -> A::PAddr {
         self.base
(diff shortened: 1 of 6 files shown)


hal-x86_64/src/lib.rs
```diff
@@ -18,9 +18,15 @@ impl Architecture for X64 {
 
 impl fmt::Debug for PAddr {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_tuple("PAddr")
-            .field(&format_args!("{:#x}", self.0))
-            .finish()
+        if let Some(width) = f.width() {
+            f.debug_tuple("PAddr")
+                .field(&format_args!("{:#0width$x}", self.0, width = width))
```
iximeow: woah til padding length can be variable

hal-core/src/mem/mod.rs
```diff
@@ -112,3 +127,19 @@ impl RegionKind {
...
 }
+
+impl fmt::Debug for RegionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            &RegionKind::FREE => f.pad("FREE"),
+            &RegionKind::USED => f.pad("USED"),
+            &RegionKind::BOOT => f.pad("BOOT"),
+            &RegionKind::BOOT_RECLAIMABLE => f.pad("BOOT_RECLAIMABLE"),
+            &RegionKind::BAD => f.pad("BAD T_T"),
```
iximeow: T_T

hal-core/src/mem/mod.rs
```diff
@@ -112,3 +127,19 @@ impl RegionKind {
     /// Memory used for a page table.
     pub const PAGE_TABLE: Self = Self(KindInner::PageTable);
 }
+
+impl fmt::Debug for RegionKind {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
```
iximeow: do you know how mmio shows up here? is this not a concern of the memory map at this point, pending further bootstrap, or is  it `USED`?

hal-x86_64/src/lib.rs
hawkw: fmt specs are surprisingly magical

hal-core/src/mem/mod.rs
hawkw: i _think_ they will show up as `Reserved` in the bootloader crate's memory map: https://github.com/rust-osdev/bootloader/blob/90f5b8910d146d6d489b70a6341d778253663cfa/src/bootinfo/memory_map.rs#L174-L175
but it isn't particularly clear from the comment. would have to actually look at the source to figure that out.

i think we'll want to get most mmio ranges from ACPI or BDA (does anyhting use that besides COM ports?) though.

PR #45, opened 2019-12-24
Small `class_enum!` tweaks

mystor:
Uses the `$v:vis` matcher to reduce some code duplication, and improves support for custom attributes.

### DIFF
--- a/pci/src/class.rs
+++ b/pci/src/class.rs
@@ -2,16 +2,21 @@ use crate::error;
 use core::convert::TryFrom;
 
 macro_rules! class_enum {
-    (pub enum $name:ident<NoProgIf> {
-        $(
-            // $($m:meta)*
-            $variant:ident = $value:expr
-        ),+
-        $(,)?
-    }) => {
+    (
+        $(#[$m:meta])*
+        $v:vis enum $name:ident<NoProgIf> {
+            $(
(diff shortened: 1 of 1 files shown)


pci/src/class.rs
```diff
@@ -2,16 +2,21 @@ use crate::error;
...
-    (pub enum $name:ident<NoProgIf> {
-        $(
-            // $($m:meta)*
-            $variant:ident = $value:expr
-        ),+
-        $(,)?
-    }) => {
+    (
+        $(#[$m:meta])*
+        $v:vis enum $name:ident<NoProgIf> {
```
hawkw: TIL that `vis` exists, that's awesome. Doesn't seem to be well documented...

pci/src/class.rs
```diff
@@ -2,16 +2,21 @@ use crate::error;
...
 macro_rules! class_enum {
-    (pub enum $name:ident<NoProgIf> {
-        $(
-            // $($m:meta)*
-            $variant:ident = $value:expr
-        ),+
-        $(,)?
-    }) => {
+    (
+        $(#[$m:meta])*
```
hawkw: I couldn't remember if you had to have a `#[...]` around the meta matcher or not (as you can tell) — thanks for fixing this!

PR #46, opened 2019-12-24
Autogenerate Display impls from doc comments

mystor:
This also includes Display implementations for `iface::Ide`, however they're quite verbose to match the [OSDev Wiki](https://wiki.osdev.org/Pci#Class_Codes).

I'm unfortunately not super satisfied with the display format. Only the first attribute on each variant is checked, and must be a doc comment. See tests for examples.

### DIFF
--- a/pci/src/class.rs
+++ b/pci/src/class.rs
@@ -1,12 +1,19 @@
 use crate::error;
 use core::convert::TryFrom;
+use core::fmt;
+
+macro_rules! replace_tt {
+    ($old:tt $new:tt) => {
+        $new
+    };
+}
 
 macro_rules! class_enum {
     (
         $(#[$m:meta])*
         $v:vis enum $name:ident<NoProgIf> {
             $(
-                $(#[$mm:meta])*
+                $(#[$($mm:tt)*])*
                 $variant:ident = $value:expr
             ),+
             $(,)?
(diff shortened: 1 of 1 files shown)


pci/src/class.rs
```diff
@@ -292,4 +413,18 @@ mod test {
             Ok(Class::MassStorage(MassStorage::Sata(iface::Sata::Achi1))),
         );
     }
+
+    #[test]
+    fn test_display() {
+        assert_eq!(
+            Class::MassStorage(MassStorage::Sata(iface::Sata::Achi1)).to_string(),
+            "Mass Storage Controller : Serial ATA : AHCI 1.0"
```
hawkw: I'd prefer to not put the spaces on both sides of the colon (which feels a little weird IMO). What do you think about maybe something like
```rust
"Mass Storage Controller: Serial ATA (AHCI 1.0)"
```
we could also consider using the alt format specifier, or the precision specifier, to control how much information is formatted?

hawkw:
@mystor shall i merge this now & we can tweak the formatting in a follow-up, or do you want to do it now?

mystor:
I'm fine with tweaking formatting in a follow-up

mystor:
@hawkw I added the changes you suggested, removing the space before the `:`, and supporting precision specifiers. Feel free to merge if you're satisfied.

PR #48, opened 2019-12-29
HAL interrupts + x86_64 interrupts

iximeow:
now that it boots and doesn't double fault i think this is in a good place to PR?

we can set x86_64 idt entries, handle them (and return without faulting 🥳), and set up the PIC so that we're not confusing interrupts, PLUS counting time and handling keystrokes both work!


.cargo/config
```diff
@@ -4,3 +4,4 @@ runner = "bootimage runner"
 [alias]
 dev-env = "install cargo-xbuild bootimage"
 run-x64 = "xrun --target=x86_64-mycelium.json"
+debug-x64 = "xrun --target=x86_64-mycelium.json -- -s -S"
```
iximeow:
```suggestion
debug-x64 = "xrun --target=x86_64-mycelium.json -- -gdb tcp::1234 -S"
```
`-s` is fine but my alias for this has `-gdb tcp::1234` so i'm less likely to make a ??? face and have to look up the flag later

hal-core/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "hal-core"
 version = "0.1.0"
-authors = ["Eliza Weisman <eliza@elizas.website>"]
```
iximeow: hey

hal-x86_64/src/cpu.rs
```diff
@@ -0,0 +1,59 @@
...
+impl Port {
+    pub const fn at(address: u16) -> Self {
+        Port { num: address }
+    }
+
+    pub unsafe fn readb(&self) -> u8 {
+        let result: u8;
+        asm!("inb %dx, %al" : "={al}"(result) : "{dx}"(self.num) :: "volatile");
+        result
+    }
```
iximeow: clippy wants a safety doc here but all i can say is `# Safety: if you use this wrong the computer might blow up or melt`

do you think `Safety: safety of these functions is entirely dependent on correct interaction with the device these port operations communicate with` is helpful?

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+    _t: PhantomData<T>,
+}
+
+#[derive(Copy, Clone)]
+#[repr(transparent)]
+pub struct PageFaultCode(u64);
+
+#[repr(C)]
+pub struct Registers {
+    pub instruction_ptr: VAddr, // TODO(eliza): add VAddr
```
iximeow: `// TODONE`?

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+#[cold]
+#[inline(never)]
+extern "x86-interrupt" fn nop(cx: Context<'_>) {}
+
+static mut TIMER: usize = 0;
+
+struct TestHandlersImpl;
+
+impl Handlers<crate::X64> for TestHandlersImpl {
+    #[inline(never)]
```
iximeow: oh why are these `inline(never)`? i'd have expected they'd be `inline(always)` even, so they get inlined into trampolines

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+    fn code_fault<C>(cx: C)
+    where
+        C: ctx::Context<Arch = crate::X64> + ctx::CodeFault,
+    {
+        let mut vga = vga::writer();
+        vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
+        writeln!(&mut vga, "code fault\n{:#?}", cx.registers()).unwrap();
+        vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
+        loop {}
```
iximeow: we need an x86 oops function that masks interrupts and sits in a `hlt` loop

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+            H::timer_tick();
+            unsafe {
+                PIC.end_interrupt(0x20);
+            }
+        }
+
+        extern "x86-interrupt" fn keyboard_isr<H: Handlers<crate::X64>>(_regs: &mut Registers) {
+            H::keyboard_controller();
+            unsafe {
+                PIC.end_interrupt(0x21);
```
iximeow: this is what i meant elsewhere about having interrupt-source-sensitive trampolines. how do you feel about having `Handlers` be hal-crate-specific bundles of traits? ARM's division by zero condition is just a flag bit, for example, where it's an interrupt for x86.

having bundles of interrupt handlers i think means we can pass some composite type through to `CascadedPic::register_handlers` with a `&mut Idt` so it can add entries at the appropriate locations but wait for `Idt` to actually do `lidt`.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -0,0 +1,419 @@
...
+    pub const OVERFLOW: Interrupt = Interrupt::new_untyped(4);
+
+    pub const BOUND_RANGE_EXCEEDED: Interrupt = Interrupt::new_untyped(5);
+
+    pub const INVALID_OPCODE: Interrupt = Interrupt::new_untyped(6);
+
+    /// A device not available exception
+    pub const DEVICE_NOT_AVAILABLE: Interrupt = Interrupt::new_untyped(7);
+
+    // TODO(eliza): can we enforce that this diverges?
```
iximeow: `Interrupt<ErrorCode, Out = !>`?

hal-x86_64/src/interrupt/mod.rs
hawkw: yeah this should go away

hal-x86_64/src/interrupt/mod.rs
hawkw: they should be `#[inline(always)]` (and actually, i think that without an inline attr, rustc will probably inline them anyway since they never get called elsewhere...); i added this attr while debugging the issue that turned out to be caused by us not remapping PICs.

hal-x86_64/src/interrupt/mod.rs
hawkw: i think this type actually needs to mostly go away

hal-x86_64/src/cpu.rs
```diff
@@ -1,10 +1,19 @@
...
+impl fmt::Debug for Port {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Port")
+            .field("num", &format_args!("{:#02x}", self.num))
+            .finish()
+    }
+}
+
 impl Port {
     pub const fn at(address: u16) -> Self {
```
hawkw: TIOLI: it seems to me that _this_ is the actual unsafe fn, rather than `readb`/`writeb`, by calling `Port::at` you have to uphold that the address is actually a port. Once you _have_ a valid port, it is safe to read/write...

hal-x86_64/src/interrupt/pic.rs
```diff
@@ -78,23 +78,39 @@ impl hal_core::interrupt::Control for CascadedPic {
 
 impl CascadedPic {
     pub(crate) unsafe fn set_irq_addresses(&mut self, primary_start: u8, secondary_start: u8) {
+        // iowait and its uses below are guidance from the osdev wiki for compatibility with "older
+        // machines". it is not entirely clear what "older machines" exactly means, or where this
+        // is or is not necessary precisely. this code happens to work in qemu without `iowait()`,
+        // but is largely untested on real hardware where this may be a concern.
+        let iowait = || { cpu::Port::at(0x80).writeb(0) };
```
hawkw: i love slash hate this comment

hal-x86_64/src/cpu.rs
iximeow: i disagree: we don't know the semantics of the port at the given address, but we _do_ know there are 65536 i/o ports for an x86 machine. components using a specific port are then responsible for not accidentally sending bytes that might, as an example, misconfigure a device that when mixed with faulty hardware start a fire. (to be clear, this is a hypothetical, but i know of uncomfortably similar real circumstances w/ devices that don't use i/o ports)

on the other hand, just saying a port number isn't a dangerous operation.

(i think of this as, essentially, passing a pointer into a struct we later read/write through, and have about the same opinion of bare mmio whenever we get to working with that too)

hawkw: hmm...i suppose that's fair: considering a rawptr as precedent, creating a rawptr is safe, but _dereferencing_ it is unsafe, so i suppose it makes for ports to be similar. nvm.

hal-core/Cargo.toml
hawkw: whoop's...we should probably add an ixi too

hal-core/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "hal-core"
 version = "0.1.0"
-authors = ["Eliza Weisman <eliza@buoyant.io>"]
+authors = ["Eliza Weisman <eliza@elizas.website>", "iximeow <git@iximeow.net>"]
```
iximeow: :heart:

iximeow:
🥳 this is now further than i've gotten w/ any osdev before

hawkw:
yayyyyyy! it is not _quite_ further than i've ever gotten before but we have accumulated significantly less technical debt than i had last time at this point. :P

Issue #49, opened 2019-12-29
have an actual HAL-exposed oops handler


iximeow:
our tentative [x86_64 double fault handler](https://github.com/hawkw/mycelium/pull/48/files#diff-a4d7ebc529bc18ebeb4d22d0229d5d4dR82-R91) prints some cpu state and hangs, but we'll probably want a higher level kernel panic to dump kernel state _and also_ architecture-specific state

hawkw:
yeah this would be good

hawkw:
if we panic in the hal, we'll hit this: https://github.com/hawkw/mycelium/blob/dd981a3efeb9f803e6671e93977e9bf30bf8e9c5/src/lib.rs#L61-L68

maybe we want to hang more debug info off of that?

PR #50, opened 2019-12-29
feat(kernel): write traces to VGA and serial

### LINKED ISSUE #38: tracing/logging in the kernel
it would be nice to have something more structured than `writeln!`, esp if we have a COM port or something to dump verbose logs on (#36). `tracing` has no_std support, but it requires liballoc, so we can't use it before we have an allocator set up. we could also use `log`, but it would be cool to be able to use _my_ logging library...

hawkw:
This PR adds support for `tracing`

The (x86_64) kernel logs traces to both the VGA buffer and to COM1, with
traces at the `INFO`, `WARN`, and `ERROR` levels going to VGA, and all
traces going to the serial port (by default). 

Right now, all the log levels are hardcoded, but I'd like to parse a boot
command line (and maybe even add fancier filtering!).

I've also added: 
 - a simple serial port driver (necessary for serial logging)
 - a port of most of `std::io` to `mycelium_util`

An example of kernel traces on VGA:
Screen Shot 2019-12-31 at 12 53 46 PM

And on serial:
Screen Shot 2019-12-31 at 12 52 27 PM

To get serial traces, you can run:
```
:; cargo run-x64 -- -serial file:$(pwd)/target/x86_64-mycelium/serial-$(/bin/date "+%Y-%m-%d-%H:%M:%S").log
```

Depends on #48.
Closes #38


hawkw:
rebased onto master now that #49 has landed. i wanna get better interfaces for stuff like indentation and colors in the vga module but can do that separately.

util/src/io/impls.rs
```diff
@@ -0,0 +1,307 @@
...
+    #[cfg(feature = "alloc")]
+    fn read_to_end(&mut self, buf: &mut Vec<u8>) -> io::Result<usize> {
+        buf.extend_from_slice(*self);
+        let len = self.len();
+        *self = &self[len..];
+        Ok(len)
+    }
+}
+
+// TODO(eliza): add
```
mystor: What is this TODO referencing adding?

util/src/io/initializer.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    /// Indicates if a buffer should be initialized.
+    #[inline]
+    pub fn should_initialize(&self) -> bool {
+        self.0
+    }
+
+    /// Initializes a buffer if necessary.
+    #[inline]
+    pub fn initialize(&self, buf: &mut [u8]) {
```
mystor: I think this deviates from the unstable `std` API, but we should probably take a `&mut [MaybeUninit<u8>]`.

util/src/io/mod.rs
```diff
@@ -0,0 +1,1089 @@
...
+    ///
+    /// # Errors
+    ///
+    /// If this function encounters any form of I/O or other error, an error
+    /// variant will be returned. If an error is returned then it must be
+    /// guaranteed that no bytes were read.
+    ///
+    /// An error of the [`ErrorKind::Interrupted`] kind is non-fatal and the read
+    /// operation should be retried if there is nothing else to do.
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize>;
```
mystor: Given that we don't need to maintain compatibility with `std`, it would be possible to make `read` take a signature more like:

```rust
fn read<'a>(&mut self, buf: &'a mut [MaybeUninit<u8>]) -> Result<(&'a mut [u8], &'a mut [MaybeUninit<u8>])>;
```

Which would explicitly encode the initialization range.
That being said, that signature is _way_ noisier, and the current behaviour in rust has to stay sound for other reasons, so we can probably get away with copying with `std` does.

util/src/io/mod.rs
```diff
@@ -0,0 +1,1089 @@
...
+        } else {
+            g.len = g.buf.len();
+            ret
+        }
+    }
+}
+
+#[cfg(feature = "alloc")]
+fn read_until<R: BufRead + ?Sized>(r: &mut R, delim: u8, buf: &mut Vec<u8>) -> Result<usize> {
+    unimplemented!("eliza: figure out memchr!!!!")
```
mystor: `core` actually contains a full-rust implementation of `memchr` (https://github.com/rust-lang/rust/blob/0fb43801368ae8b5931583f813071120bed55c35/src/libcore/slice/memchr.rs#L40-L83) to optimize the `[_]::contains` method for `u8` and `i8` (Thanks again @Manishearth!)

Unfortunately I don't think the full functionality is exposed anywhere stably, though we could copy the impl from libcore.

util/src/io/util.rs
```diff
@@ -0,0 +1,216 @@
...
+/// `write` returns an error. All instances of `ErrorKind::Interrupted` are
+/// handled by this function and the underlying operation is retried.
+pub fn copy<R: ?Sized, W: ?Sized>(reader: &mut R, writer: &mut W) -> io::Result<u64>
+where
+    R: Read,
+    W: Write,
+{
+    let mut buf = MaybeUninit::<[u8; super::DEFAULT_BUF_SIZE]>::uninit();
+    // FIXME(eliza): the stdlib has a scary comment that says we haven't decided
+    // if this is okay or not...
```
mystor: Yeah, this has to do with the initialization thing I mention in an earlier comment. The `read` method takes a `&mut [u8]` for historical reasons, and IIRC it hasn't been decided yet whether or not passing uninitialized memory as a `&mut [u8]` is OK.

The `Initializer` mechanism is effectively being used here as a way of encoding a property of a method not encoded by the type system.

hal-x86_64/src/log.rs
```diff
@@ -0,0 +1,144 @@
+use crate::{serial, vga};
+use log::{Level, LevelFilter, Log, Metadata, Record};
+
+// #[derive(Debug)]
```
mystor: We can probably drop this derive comment.

hal-x86_64/src/log.rs
```diff
@@ -0,0 +1,144 @@
...
+impl Log for Logger {
+    fn enabled(&self, metadata: &Metadata) -> bool {
+        let level = metadata.level();
+        self.vga_enabled(level) || self.serial_enabled(level)
+    }
+
+    fn log(&self, record: &Record) {
+        let level = record.level();
+        if self.vga_enabled(level) {
+            use core::fmt::Write;
```
mystor: I'm thinking we'll probably want to make `vga::writer` use `mycelium_util::io::Write` as well.

hal-x86_64/src/log.rs
```diff
@@ -0,0 +1,144 @@
...
+            let mut vga = vga::writer();
```
mystor: In the case of these special types which are intended to be used with `write!`-style macros, we may want to add an inherent `write_fmt` method, like what `core` does with `Formatter`. It allows you to avoid importing the `Write` trait to write into the type. https://github.com/rust-lang/rust/blob/0fb43801368ae8b5931583f813071120bed55c35/src/libcore/fmt/mod.rs#L1468-L1489

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -136,22 +132,14 @@ impl Handlers<crate::X64> for TestHandlersImpl {
...
         let mut vga = vga::writer();
         vga.set_color(vga::ColorSpec::new(vga::Color::Yellow, vga::Color::Black));
-        writeln!(
-            &mut vga,
-            "lol im in ur test interrupt\n{:#?}",
-            cx.registers()
-        )
-        .unwrap();
+        log::info!("lol im in ur test interrupt\n{:#?}", cx.registers());
         vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
```
mystor: More `set_color` calls.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -58,7 +58,7 @@ impl Handlers<crate::X64> for TestHandlersImpl {
     {
         let mut vga = vga::writer();
         vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
-        writeln!(&mut vga, "page fault\n{:#?}", cx.registers()).unwrap();
+        log::error!("page fault\n{:#?}", cx.registers());
         vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
```
mystor: Given that the Logger now handles adding `[LEVEL]` tags & doing colouring, we can probably get rid of the call to `vga::writer()` and the calls to `vga.set_color` entirely.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -69,7 +69,7 @@ impl Handlers<crate::X64> for TestHandlersImpl {
     {
         let mut vga = vga::writer();
         vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
-        writeln!(&mut vga, "code fault\n{:#?}", cx.registers()).unwrap();
+        log::error!("code fault\n{:#?}", cx.registers());
         vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
```
mystor: Same deal with removing `set_color` calls.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -81,7 +81,7 @@ impl Handlers<crate::X64> for TestHandlersImpl {
     {
         let mut vga = vga::writer();
         vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
-        writeln!(&mut vga, "double fault\n{:#?}", cx.registers()).unwrap();
+        log::error!("double fault\n{:#?}", cx.registers());
         vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
```
mystor: More `set_color` calls.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -119,13 +116,12 @@ impl Handlers<crate::X64> for TestHandlersImpl {
...
-            &mut vga,
+        log::info!(
+            // for now
             "got scancode {}. the time is now: {}",
             scancode,
             unsafe { TIMER }
-        )
-        .unwrap();
+        );
         vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
```
mystor: More `set_color` calls.

util/src/io/mod.rs
```diff
@@ -0,0 +1,1089 @@
...
+        if g.len == g.buf.len() {
+            unsafe {
+                // FIXME(danielhenrymantilla): #42788
+                //
+                //   - This creates a (mut) reference to a slice of
+                //     _uninitialized_ integers, which is **undefined behavior**
+                //
+                //   - Only the standard library gets to soundly "ignore" this,
+                //     based on its privileged knowledge of unstable rustc
+                //     internals;
```
mystor: Another comment related to the ones above talking about not being sure if the calls to `initialize` are "ok"

util/src/io/error.rs
```diff
@@ -0,0 +1,97 @@
+pub type Result<T> = core::result::Result<T, self::Error>;
+#[derive(Debug)]
+pub struct Error {
+    kind: ErrorKind,
+    details: Option<&'static str>,
+}
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
+pub enum ErrorKind {
```
mystor: I'm guessing we aren't using most of these `ErrorKind` values right now, and it's possible some will never be used in a kernel (e.g. `PermissionDenied`).
It might be worth trimming down this list a bit to only contain values we need in a kernel, and add new ones as we need them, don't have strong feelings about that, though.

util/src/lib.rs
```diff
@@ -2,6 +2,10 @@
 //! libraries.
 #![cfg_attr(target_os = "none", no_std)]
 
+#[cfg(feature = "alloc")]
+extern crate alloc;
```
mystor: I'm guessing this feature will be enabled by our allocator crate when it is added as a dependency?

Cargo.toml
```diff
@@ -16,6 +16,7 @@ name = "mycelium_kernel"
 
 [dependencies]
 hal-core = { path = "hal-core" }
+log = { version = "0.4", default_features = false }
```
mystor: It might be worth re-exporting the `log` macros from `mycelium_util`, so we don't need to remember to set `default_features = false` on every crate where we want to use logging macros. Not sure though.

hawkw: not a bad idea!

util/src/io/impls.rs
hawkw: there was a commented-out `BufRead` impl before i added `BufRead`; this should be removed now

util/src/lib.rs
hawkw: it's enabled by the kernel when it has an allocator and/or by code which requires alloc.

hal-x86_64/src/interrupt/mod.rs
hawkw: yeah, i did that in my follow-up PR that makes the interrupt handlers more cross-platform... probably should've done it here instead

hawkw:
UPDATE: now that we have an allocator (thanks @mystor!) we can now use `tracing` rather than `log`!

util/src/io/error.rs
hawkw: I was going to do this initially, but I kept needing to add more error kinds. I am also anticipating we _might_ want to use mycelium_util as a replacement for std in non-kernel-mode components (microkernel servers?) until we can target std at mycelium's interface?

Cargo.toml
hawkw: hmm the downside to this is that the macros are now imported from `mycelium_util`, which feels a little weird...i like how with `log::info!`/`tracing::info!`, the crate name describes what the macro does...

hawkw:
note that some of the formatting code in the subscriber impl is pretty complicated...this is mostly because i wanted slightly different formatting on VGA and serial (since the VGA buffer is only 80 columns wide, i've tried to keep the line length down)

util/src/io/mod.rs
hawkw: IMHO `[MaybeUninit<u8>]` is a better approach, but duplicating std exactly lets us copy a lot more code from std without modifying it...I'm open to changing this in a follow-up.

hal-core/src/boot.rs
```diff
@@ -18,6 +18,10 @@ pub trait BootInfo {
         None
     }
 
+    fn subscriber(&self) -> Option<tracing::Dispatch> {
+        None
+    }
+
     // TODO(eliza): figure out a non-bad way to represent boot command lines (is
     // it reasonable to convert them to rust strs when we barely have an operating
     // system???)
```
mystor: Completely unrelated, but I think it'd be reasonable to do something like that. We can panic early during boot if a boot command line isn't valid utf-8 or something.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -57,21 +57,15 @@ impl Handlers<crate::X64> for TestHandlersImpl {
...
     fn code_fault<C>(cx: C)
     where
         C: ctx::Context<Arch = crate::X64> + ctx::CodeFault,
     {
-        let mut vga = vga::writer();
-        vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
-        writeln!(&mut vga, "code fault\n{:#?}", cx.registers()).unwrap();
-        vga.set_color(vga::ColorSpec::new(vga::Color::Green, vga::Color::Black));
+        tracing::error!(registers = ?cx.registers(), "code fault");
         loop {}
```
mystor: I suppose we could probably `panic!` here and in the other fatal interrupts as well, rather than `loop {}`-ing.

util/src/io/impls.rs
mystor: This is still around fwiw

hal-x86_64/src/interrupt/mod.rs
hawkw: in a follow-up, i am planning on making them call into a "kernel oops" handler that the panic handler also calls into, but with different args. in the future, some of these will at least occasionally be recoverable (in user code)

PR #52, opened 2019-12-31
feat(alloc): add a super basic and limited allocator

mystor:
I thought it might be useful to add a dummy kernel allocator, in case it is useful for building other components, before a proper paging abstraction is added to the `hal`. This allocator is about as dumb as it gets, providing only 640k of bump-allocating power, but allows the use of types like `Vec` in the kernel.

### DIFF
--- a/alloc/src/lib.rs
+++ b/alloc/src/lib.rs
@@ -1,0 +1,112 @@
+//! The simplest thing resembling an "allocator" I could possibly create.
+//! Allocates into a "large" static array.
+#![no_std]
+
+use core::alloc::Layout;
+use core::cell::UnsafeCell;
+use core::mem::MaybeUninit;
+use core::ptr;
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+#[cfg(target_os = "none")]
+use core::alloc::GlobalAlloc;
+
+// 640k is enough for anyone
+const HEAP_SIZE: usize = 640 * 1024;
+
+#[repr(align(16))]
+struct Heap(UnsafeCell<MaybeUninit<[u8; HEAP_SIZE]>>);
+
+unsafe impl Sync for Heap {}
+
+static HEAP: Heap = Heap(UnsafeCell::new(MaybeUninit::uninit()));
+static USED: AtomicUsize = AtomicUsize::new(0);
+
+/// NOTABLE INVARIANTS:
+///  * `Layout` is non-zero sized (enforced by `GlobalAlloc`)
+///  * `align` is a power of two (enforced by `Layout::from_size_align`)
+pub unsafe fn alloc(layout: Layout) -> *mut u8 {
+    let heap = HEAP.0.get() as *mut u8;
+
+    let mut prev = USED.load(Ordering::Relaxed);
+    loop {
+        let align_offset = heap.add(prev).align_offset(layout.align());
+
+        // NOTE: layout.size() is non-zero. The check will fail if
+        // `align_offset` exceeds `HEAP_SIZE - prev`.
+        let space = (HEAP_SIZE - prev).saturating_sub(align_offset);
+        if space < layout.size() {
+            return ptr::null_mut();
+        }
+
+        let next = prev + align_offset + layout.size();
+        match USED.compare_exchange_weak(prev, next, Ordering::Relaxed, Ordering::Relaxed) {
+            Ok(_) => {
+                return heap.add(prev + align_offset);
+            }
+            Err(next_prev) => {
+                prev = next_prev;
+                continue;
+            }
+        }
+    }
+}
+
+pub unsafe fn dealloc(_ptr: *mut u8, _layout: Layout) {
+    // lol
+}
+
+#[cfg(target_os = "none")]
+pub struct Global;
+
+#[cfg(target_os = "none")]
+unsafe impl GlobalAlloc for Global {
+    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
+        alloc(layout)
+    }
+
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
+        dealloc(ptr, layout)
+    }
+}
+
+#[cfg(target_os = "none")]
+#[global_allocator]
+pub static GlOBAL: Global = Global;
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use core::mem;
+
+    #[test]
+    fn alloc_small() {
+        let p0 = unsafe { alloc(Layout::new::<u32>()) };
+        assert!(!p0.is_null());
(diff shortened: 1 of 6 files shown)


alloc/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+use core::alloc::Layout;
+use core::cell::UnsafeCell;
+use core::mem::MaybeUninit;
+use core::ptr;
+use core::sync::atomic::{AtomicUsize, Ordering};
+
+#[cfg(target_os = "none")]
+use core::alloc::GlobalAlloc;
+
+// 640k is enough for anyone
```
hawkw: :D

alloc/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+
+        // NOTE: layout.size() is non-zero. The check will fail if
+        // `align_offset` exceeds `HEAP_SIZE - prev`.
+        let space = (HEAP_SIZE - prev).saturating_sub(align_offset);
+        if space < layout.size() {
+            return ptr::null_mut();
+        }
+
+        let next = prev + align_offset + layout.size();
+        match USED.compare_exchange_weak(prev, next, Ordering::Relaxed, Ordering::Relaxed) {
```
hawkw: "Relaxed, Relaxed" kinda scares me — at a glance, it seems like there should be an acquire/release pair guarding the actual handing out of ptrs, which mustn't be reordered?

alloc/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+            Err(next_prev) => {
+                prev = next_prev;
+                continue;
+            }
+        }
+    }
+}
+
+pub unsafe fn dealloc(_ptr: *mut u8, _layout: Layout) {
+    // lol
```
hawkw: lol

src/lib.rs
```diff
@@ -1,8 +1,17 @@
 #![cfg_attr(target_os = "none", no_std)]
+#![cfg_attr(target_os = "none", feature(alloc_error_handler))]
+
+extern crate alloc;
+
+// Force `mycelium_alloc` to be linked in, as it provides the allocator.
+#[cfg(target_os = "none")]
+extern crate mycelium_alloc;
```
hawkw: Ah, interesting...I took the approach of just defining a bunch of allocator implementations in the alloc crate, and defining the global allocator in the kernel itself, so we can (eventually) have more policy in the kernel (like a tiered scheme where we switch the allocator once paging is set up).

src/lib.rs
```diff
@@ -50,6 +59,18 @@ where
...
+    {
+        // Let's allocate something, for funsies
+        let mut v = Vec::new();
+        writeln!(&mut writer, "vec: {:?} (@ {:p})", v, v.as_ptr()).unwrap();
+        v.push(5u64);
+        writeln!(&mut writer, "vec: {:?} (@ {:p})", v, v.as_ptr()).unwrap();
+        v.push(10u64);
+        writeln!(&mut writer, "vec: {:?} (@ {:p})", v, v.as_ptr()).unwrap();
+        assert_eq!(v.pop(), Some(10));
+        assert_eq!(v.pop(), Some(5));
```
hawkw: awesome how this actually "works" :D

src/arch/x86_64.rs
```diff
@@ -73,3 +73,11 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
...
     mycelium_kernel::handle_panic(&mut vga, info)
 }
+
+#[alloc_error_handler]
+#[cfg(target_os = "none")]
+fn alloc_error(layout: core::alloc::Layout) -> ! {
+    let mut vga = vga::writer();
+    vga.set_color(vga::ColorSpec::new(vga::Color::Red, vga::Color::Black));
+    mycelium_kernel::handle_alloc_error(&mut vga, layout)
+}
```
hawkw: Take it or leave it: this could also just panic, which could be platform-agnostic. Unless we eventually want to do Real Stuff here.

hawkw:
btw, @mystor, have you seen @fitzgen's [post](https://fitzgeraldnick.com/2019/11/01/always-bump-downwards.html) on bumping "downwards" instead of "upwards" to avoid unnecessary overflow checks?

alloc/src/lib.rs
mystor: I might be wrong, but I _think_ it's OK because we never free memory.
There are no writes to other memory regions than the atomic which need to be made visible to other allocating threads, as each allocation region is fully disjoint, so we just need to ensure that reads and writes to the `USED` atomic are consistent, which `Relaxed` handles.

src/lib.rs
mystor: Ah, that's also super reasonable. I just threw it into the `alloc` crate 'cause it was easy to do :p

src/arch/x86_64.rs
mystor: True! It can panic as we're not unwinding.
I just put it next to the panic handler 'cause I was cargo-culting everything not directly related to the allocator tbh.

mystor:
> btw, @mystor, have you seen @fitzgen's [post](https://fitzgeraldnick.com/2019/11/01/always-bump-downwards.html) on bumping "downwards" instead of "upwards" to avoid unnecessary overflow checks?

I hadn't, but I'm _kinda_ already doing that in this PR :-P
The bounds checks for the allocation are all done "downward", and then the final pointer math is done upward, in order to somewhat avoid overflow checks (though there is still one underflow check due to the need to combine alignment correction & the allocation size).
https://github.com/hawkw/mycelium/pull/52/files#diff-9692300bf796849339912751061a99bfR35-R40

I'll push an update which properly bumps downwards.

alloc/src/lib.rs
hawkw: ah, hmm, yeah i think you're right!

hawkw: ah, hmm, yeah i think you're right!

hawkw:
> I hadn't, but I'm _kinda_ already doing that in this PR :-P

yeah, i noticed! that's why i thought you'd dig it.

alloc/src/lib.rs
hawkw: also it doesn't actually matter, because by the time we know about "cores" we will probably have switched over to an allocation strategy that supports "deallocating"

mystor: "deallocating"?! never heard of it :stuck_out_tongue:

hawkw:
happy to merge this PR whenever you feel good about it — we can make the allocation strategy fancier later!

hawkw:
happy to merge this PR whenever you feel good about it — we can make the allocation strategy fancier later!

PR #54, opened 2019-12-31
avoid deadlocks during timer interrupts

mystor:


hawkw:
i think i accidentally clobbered this in a different PR (perhaps the logging one?)

PR #55, opened 2020-01-01
feat(kernel): dedicated kernel oops handler

### LINKED ISSUE #49: have an actual HAL-exposed oops handler

hawkw:
This branch adds a (basic) kernel oops handler for when we do a
fucky-wucky. Right now it just makes a cool red screen of death 
and dumps a couple control registers, there's probably a lot of
additional state we should dump in here in the future.

Panics are also logged to serial prior to oopsing.

Right now it looks like this:
Screen Shot 2019-12-31 at 2 44 42 PM

Fatal CPU exception ISRs should eventually call into oops directly,
rather than by panicking, so we can dump additional exception info.

Closes #49 (I'm assuming we will open follow-up issues for getting
more useful info on the oops screen).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,7 +4,6 @@ members = [
     "util",
     "hal-core",
     "hal-x86_64",
-    "hal-multiboot2",
 ]
 
 [package]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
     "util",
     "hal-core",
     "hal-x86_64",
+    "hal-multiboot2",
 ]
 
 [package]
--- a/hal-x86_64/src/vga.rs
+++ b/hal-x86_64/src/vga.rs
@@ -5,8 +5,6 @@ lazy_static::lazy_static! {
     static ref BUFFER: spin::Mutex<Buffer> = spin::Mutex::new(Buffer {
         col: 0,
         row: 0,
-        indent: 0,
-        chars_this_line: false,
         color: ColorSpec::new(Color::LightGray, Color::Black),
         buf: unsafe { &mut *(0xb8000 as *mut Buf) },
     });
@@ -47,8 +45,6 @@ pub struct ColorSpec(u8);
 pub struct Buffer {
     col: usize,
     row: usize,
-    indent: usize,
-    chars_this_line: bool,
     color: ColorSpec,
     buf: &'static mut Buf,
 }
@@ -121,10 +117,6 @@ impl Buffer {
 
         self.buf[self.row][self.col].write(self.character(ch));
         self.col += 1;
-
-        if ch != b' ' {
-            self.chars_this_line = true;
-        }
     }
 
     fn newline(&mut self) {
@@ -146,8 +138,18 @@ impl Buffer {
             c.write(blank)
         }
 
-        self.col = self.indent;
-        self.chars_this_line = false;
+        self.col = 0;
+    }
+
+    fn clear(&mut self) {
+        let ch = self.character(b' ');
+        for row in self.buf.iter_mut() {
+            for col in row.iter_mut() {
+                col.write(ch);
+            }
+        }
+        self.row = 0;
+        self.col = 0;
     }
 }
 
@@ -174,26 +176,8 @@ impl Writer {
         BUFFER.lock().set_color(color);
     }
 
-    pub fn indent(&mut self, indent: u32) {
-        self.dent(indent as isize)
-    }
-
-    pub fn outdent(&mut self, outdent: u32) {
-        self.dent(-(outdent as isize))
-    }
-
-    /// Indent or outdent the start of each line by the requested number of
-    /// characters.
-    pub fn dent(&mut self, dent: isize) {
-        let mut lock = BUFFER.lock();
-        if dent > 0 {
-            lock.indent = lock.indent.saturating_add(dent as usize);
-        } else {
-            lock.indent = lock.indent.saturating_sub((-dent) as usize)
-        }
-        if !lock.chars_this_line {
-            lock.col = lock.indent;
-        }
+    pub fn clear(&mut self) {
+        BUFFER.lock().clear();
     }
 }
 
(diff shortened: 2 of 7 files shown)


src/arch/x86_64.rs
```diff
@@ -72,10 +72,39 @@ pub extern "C" fn _start(info: &'static bootinfo::BootInfo) -> ! {
...
+    };
+    let _ = writeln!(
+        vga,
+        "\n  cr0: {:#032b}\n  cr3: {:#032b}\n  rflags: {:#064b}",
+        cr0, cr3, rflags
+    );
+    let _ = vga.write_str("\n\n  it will never be safe to turn off your computer.");
+
+    #[allow(clippy::empty_loop)]
+    loop {}
```
mystor: It might be worthwhile to halt the CPU, in addition to disabling interrupts, when we panic. Something like `loop { unsafe { asm!("hlt" :::: "volatile"); } }` might do it?

src/arch/x86_64.rs
```diff
@@ -72,10 +72,39 @@ pub extern "C" fn _start(info: &'static bootinfo::BootInfo) -> ! {
     mycelium_kernel::kernel_main(&bootinfo);
 }
 
-#[panic_handler]
-#[cfg(target_os = "none")]
-fn panic(info: &core::panic::PanicInfo) -> ! {
+pub(crate) fn oops(cause: &dyn core::fmt::Display) -> ! {
+    use core::fmt::Write;
+
+    unsafe { asm!("cli" :::: "volatile") }
```
mystor: Should we also stub out the IDT to prevent NMIs from being run?

src/arch/x86_64.rs
```diff
@@ -72,10 +72,39 @@ pub extern "C" fn _start(info: &'static bootinfo::BootInfo) -> ! {
...
     let mut vga = vga::writer();
```
mystor: We should probably avoid trying to lock the VGA lock during the oops handler, which I think we do in methods like `set_color`, in case they are locked when the panic occurs.

It'd probably be reasonable (at least for now, as we don't have "threads") to attempt to take the lock and unsafely bypass it if we can't.

src/lib.rs
```diff
@@ -68,11 +68,13 @@ where
         v.push(5u64);
         tracing::info!(vec = ?v, vec.addr = ?v.as_ptr());
         v.push(10u64);
-        tracing::info!(vec=?v, vec.addr = ?v.as_ptr());
+        tracing::info!(vec=?v, vec.addr=?v.as_ptr());
         assert_eq!(v.pop(), Some(10));
         assert_eq!(v.pop(), Some(5));
     }
 
+    panic!("fake panic!");
```
mystor: We prob. want to remove this before we merge

src/arch/mod.rs
```diff
@@ -1,2 +1,5 @@
 #[cfg(target_arch = "x86_64")]
 pub(crate) mod x86_64;
+
+#[cfg(target_arch = "x86_64")]
+pub(crate) use self::x86_64::oops;
```
mystor: Perhaps we should do a `pub(crate) use self::x86_64::*;` instead, and refer to things which should be provided by any arch as `crate::arch::WHATEVER`.
It would let us avoid `target_arch` checks in other parts of our code.

hal-x86_64/src/vga.rs
```diff
@@ -146,8 +138,18 @@ impl Buffer {
...
+        self.col = 0;
+    }
+
+    fn clear(&mut self) {
+        let ch = self.character(b' ');
+        for row in self.buf.iter_mut() {
+            for col in row.iter_mut() {
+                col.write(ch);
+            }
+        }
```
mystor: It might be more efficient to do a single volatile write to the entire VGA buffer, if that's allowed, like:

```rust
ptr::write_volatile(&mut *self.buf, [[self.character(b' '); BUF_WIDTH]; BUF_HEIGHT]);
```

src/arch/x86_64.rs
iximeow: alternate very cursed idea: swap IDTs so irq0 is the real oops handler, then do an `int 0` to raise the interrupt, thereby blocking any other NMI until the handler returns

src/lib.rs
hawkw: oopsie-whoopsie!

hal-x86_64/src/vga.rs
hawkw: hmm, i wonder what that compiles to...

src/arch/x86_64.rs
hawkw: hmm, yeah...i didn't initially want to add this, but i think it makes sense for our mutex to have an `unsafe fn force_unlock` or something specifically for cases where we _know_ a previously acquired lock will never be used again (basically when we're oopsing). even when we are running on multiple cores, i think we would want to kill them pretty early in the oops process...

src/arch/x86_64.rs
hawkw: > Should we also stub out the IDT to prevent NMIs from being run?

hmm, if we just make the NMI handler "not present", we'll fault. i think we would either need a nop ISR, or do ixi's horrible thing.

hawkw: i think if an NMI happens during an oops, that's real bad news indeed...

src/arch/x86_64.rs
mystor: Yeah. I imagine we want to halt all cores before interacting with VGA at all, at which point forcing unlocks should be "safe".

src/arch/x86_64.rs
mystor: I think either ixi's idea or installing a nop-ISR are the main options here.

Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = [
     "util",
     "hal-core",
     "hal-x86_64",
+    "hal-multiboot2",
```
mystor: I'm guessing this got added to the commit accidentally?

hawkw: asdf

PR #58, opened 2020-01-03
feat(wasm): run a "hello world" wasm under an interpreter

mystor:
This uses a fork of the `wasmi` crate as the interpreter: https://github.com/mystor/wasmi/tree/mycelium.
This fork has been modified to run under our custom target, removing implicit `std` dependencies, and working around some missing soft-float intrinsics.


hawkw:
holy cow

src/lib.rs
```diff
@@ -8,6 +8,10 @@ use hal_core::{boot::BootInfo, mem, Architecture};
 
 use alloc::vec::Vec;
 
+mod wasm;
+
+const HELLOWORLD_WASM: &[u8] = include_bytes!("helloworld.wasm");
```
hawkw: :D :D :D :D

src/lib.rs
```diff
@@ -8,6 +8,10 @@ use hal_core::{boot::BootInfo, mem, Architecture};
...
+
 pub fn kernel_main<A>(bootinfo: &impl BootInfo<Arch = A>) -> !
```
hawkw: this clippy lint fires a lot of false positives on tracing macros, we should just turn it off

src/wasm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+                        }
+                    ),*
+                }
+            }
+        }
+    }
+}
+
+host_funcs! {
+    fn "wasi_unstable"::"fd_write"(fd: u32, iovs: u32, iovs_len: u32, nwritten: u32) -> u16
```
hawkw: TIOLI: seems like these could be idents rather than quoted?

src/wasm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        };
+
+        Ok(Host {
+            module: instance.clone(),
+            memory,
+        })
+    }
+}
+
+macro_rules! host_funcs {
```
hawkw: am i correct that this macro is generating all the wasi host calls (eventually)? it would be nice to have a little more documentation of all the stuff this generates & how it should be invoked (eventually)

src/wasm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+    ($(
+        fn $module:literal :: $name:literal ($($p:ident : $t:ident),*) $( -> $rt:ident)?
```
hawkw: whoa, TIL that `literal` is an acceptable thing to match against in a macro!

src/wasm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+            }
+        }
+
+        impl wasmi::Externals for Host {
+            fn invoke_index(
+                &mut self,
+                index: usize,
+                args: wasmi::RuntimeArgs,
+            ) -> Result<Option<wasmi::RuntimeValue>, wasmi::Trap> {
+                let span = tracing::trace_span!("invoke_index", index, ?args);
```
hawkw: TIOLI: seems like we could just slap a `#[tracing::instrument]` on the generated function?

src/wasm/wasi.rs
```diff
@@ -0,0 +1,103 @@
+use super::Host;
+
+// FIXME: These should both be `u16`, and probably generated from the wasi
+// snapshot_0 `witx` definitions.
```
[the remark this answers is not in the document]
mystor: Oops, yeah. I fixed that. Should probably remove this comment.

src/wasm/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        descriptor: &wasmi::GlobalDescriptor,
+    ) -> Result<wasmi::GlobalRef, wasmi::Error> {
+        tracing::error!(
+            module_name,
+            field_name,
+            ?descriptor,
+            "unresolved global import"
+        );
+        Err(wasmi::Error::Instantiation(
+            "unresolved global import".to_owned(),
```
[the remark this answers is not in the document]
mystor: Yup :-/. Every `wasmi` error has a string allocation.
if we decide to hard-fork `wasmi` and break the API, we could change the error type (and the lib in general) to be much less allocation-happy.

src/wasm/mod.rs
mystor: *shrug* probably.
I made them strings because I think _theoretically_ wasm supports spaces in symbols. I'm not sure we'll ever declare a symbol which has spaces in it, though, so it's probably not reasonable.

src/wasm/mod.rs
mystor: Yeah, it's super handy, especially because it has an open FOLLOW set, so you don't need to be as careful with what you put after it!
Been around since ~2016 sometime, but was stabilized in https://github.com/rust-lang/rust/pull/56072.

src/wasm/mod.rs
mystor: This is the method which I was poking you on Discord about yesterday which breaks when annotated with `#[tracing::instrument]` due to rust messing up spans.

src/wasm/mod.rs
mystor: Yeah. I threw together this macro because I needed to write a boatload of boilerplate to declare even a single host method, and didn't want to have to do it again.

We might actually want to use `witx` files and auto-generate the binding logic (similar to how `wasmtime` does since https://github.com/bytecodealliance/wasmtime/pull/707), but this was much easier to get started with.

src/wasm/mod.rs
hawkw: ah gotcha. it would be nice to have a tracing issue for that (even though it's a compiler error), just so we can reference the related issue against the compiler?

src/helloworld.wast
```diff
@@ -0,0 +1,32 @@
+;; Copied from the wasmtime WASI tutorial
+;; https://github.com/bytecodealliance/wasmtime/blob/1f8921ef09667dc4319c1b65ebb1ce993e19badf/docs/WASI-tutorial.md#web-assembly-text-example
+;;
+;; wat2wasm helloworld.wast
+
+(module
```
iximeow: could we wat2wasm when building the kernel? it seems super easy to accidentally end up with a binary wasm dep in the kernel that doesn't match some associated source

src/lib.rs
iximeow: :heavy_plus_sign: :heavy_plus_sign:

src/helloworld.wast
mystor: That'd be pretty solid. We could also use a pure-rust wat2wasm converter (there are a few iirc) and just run it in our build script.
I just did it this way because it was the easiest, but not checking in binaries is probably a good plan.

PR #60, opened 2020-01-14
refac(hal): do away with the Architecture trait

hawkw:
The `Architecture` trait means that pretty much all downstream code that
consumes `hal-core` has to be parameterized over an architecture type,
which is kind of painful. `Architecture` pretty much only provides two
things: definitions of the `PAddr` and `VAddr` types, and an API
contract for architecture implementations.

On most architectures that we are likely to want to target, it's pretty
reasonable to represent `PAddr`s and `VAddr`s as words instead, using
`usize` as the inner repr. This means that for most architectures, HAL
implementations can still be written in separate crates without changing
`hal-core`. For architectures where we _do_ wish to add
architecture-specific logic in the address types, such as validating
that x86 addresses are in canonical form, or which require `PAddr`s or
`VAddr`s to have non-`usize` reprss, we can add `#[cfg(target_arch =
...)]` versions of the address types in `hal-core`.

As far as providing a contract for the shape of the top-level arch APIs,
we can probably solve this through documentation, if there are ever
third-party HAL impls. This may also not matter as much as I thought it
might, since non-arch crates written against the HAL will still probably
be generic over `hal-core` traits for whatever specific HAL APIs they
depend on.

I've also moved the interrupt handler implementations out of the 
`hal-x86_64` crate and into the kernel proper!

### DIFF
--- a/hal-core/src/addr.rs
+++ b/hal-core/src/addr.rs
@@ -1,0 +1,262 @@
+use core::{fmt, ops};
+
+pub trait Address:
+    Copy
+    + ops::Add<usize, Output = Self>
+    + ops::Sub<usize, Output = Self>
+    + ops::AddAssign<usize>
+    + ops::SubAssign<usize>
+    + PartialEq
+    + Eq
+    + PartialOrd
+    + Ord
+    + fmt::Debug
+{
+    fn as_usize(self) -> usize;
+
+    /// Aligns `self` up to `align`.
+    ///
+    /// The specified alignment must be a power of two.
+    fn align_up<A: Into<usize>>(self, align: A) -> Self;
+
+    /// Aligns `self` down to `align`.
+    ///
+    /// The specified alignment must be a power of two.
+    fn align_down<A: Into<usize>>(self, align: A) -> Self;
+
+    /// Offsets this address by `offset`.
+    ///
+    /// If the specified offset would overflow, this function saturates instead.
+    fn offset(self, offset: i32) -> Self {
+        if offset > 0 {
+            self + offset as usize
+        } else {
+            let offset = -offset;
+            self - offset as usize
+        }
+    }
+
+    /// Returns the difference between `self` and `other`.
+    fn difference(self, other: Self) -> isize {
+        if self > other {
+            -(self.as_usize() as isize - other.as_usize() as isize)
+        } else {
+            (other.as_usize() - self.as_usize()) as isize
+        }
+    }
+
+    /// Returns `true` if `self` is aligned on the specified alignment.
+    fn is_aligned<A: Into<usize>>(self, align: A) -> bool {
+        self.align_down(align) == self
+    }
+
+    fn as_ptr(&self) -> *const ();
+}
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(transparent)]
+pub struct PAddr(usize);
+
+#[derive(Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+#[repr(transparent)]
+pub struct VAddr(usize);
+
+impl fmt::Debug for PAddr {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(width) = f.width() {
+            f.debug_tuple("PAddr")
+                .field(&format_args!("{:#0width$x}", self.0, width = width))
+                .finish()
+        } else {
+            f.debug_tuple("PAddr")
+                .field(&format_args!("{:#x}", self.0,))
+                .finish()
+        }
+    }
+}
+
+impl ops::Add<usize> for PAddr {
+    type Output = Self;
+    fn add(self, rhs: usize) -> Self {
+        PAddr(self.0 + rhs)
+    }
+}
+
+impl ops::Add for PAddr {
+    type Output = Self;
+    fn add(self, rhs: Self) -> Self {
+        PAddr(self.0 + rhs.0)
+    }
+}
+
+impl ops::AddAssign for PAddr {
+    fn add_assign(&mut self, rhs: Self) {
+        self.0 += rhs.0;
+    }
+}
+
+impl ops::AddAssign<usize> for PAddr {
+    fn add_assign(&mut self, rhs: usize) {
+        self.0 += rhs;
+    }
+}
+
+impl ops::Sub<usize> for PAddr {
+    type Output = Self;
+    fn sub(self, rhs: usize) -> Self {
+        PAddr(self.0 - rhs)
+    }
+}
+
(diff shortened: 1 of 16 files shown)


hal-core/src/addr.rs
```diff
@@ -0,0 +1,262 @@
...
+            (other.as_usize() - self.as_usize()) as isize
+        }
+    }
+
+    /// Returns `true` if `self` is aligned on the specified alignment.
+    fn is_aligned<A: Into<usize>>(self, align: A) -> bool {
+        self.align_down(align) == self
+    }
+
+    fn as_ptr(&self) -> *const ();
```
mystor: I wonder if it'd be nice to do `as_ptr<T>(&self) -> *const T`?

Out of curiosity, is there a particular reason we're using `*const T` instead of `*mut T`? Because `*mut` can implicitly downcast, I find methods returning it in unsafe code a bit nicer to work with.

hal-core/src/addr.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn sub_assign(&mut self, rhs: Self) {
+        self.0 -= rhs.0;
+    }
+}
+
+impl ops::SubAssign<usize> for PAddr {
+    fn sub_assign(&mut self, rhs: usize) {
+        self.0 -= rhs;
+    }
+}
```
mystor: Should we consider pulling in a dependency for auto-deriving these sort of methods on newtypes?

hal-core/src/addr.rs
```diff
@@ -0,0 +1,262 @@
...
+
+impl Address for PAddr {
+    fn as_usize(self) -> usize {
+        self.0 as usize
```
mystor: Isn't `self.0` already a usize?

hal-core/src/addr.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        if let Some(width) = f.width() {
+            f.debug_tuple("VAddr")
+                .field(&format_args!("{:#0width$x}", self.0, width = width))
+                .finish()
+        } else {
+            f.debug_tuple("VAddr")
+                .field(&format_args!("{:#x}", self.0,))
+                .finish()
+        }
```
mystor: I wish that you could pass an `Option<usize>` into `width = ` with `format_args!`. Would make stuff like this much nicer to read.

If we feel like being cheeky here, I think it's possible to write `fmt::LowerHex::fmt(&self.0, f)` to get the field printed, but we'd need to write out the `"VAddr(...)"` ourselves.

I could also imagine always wanting to use a full-width printout for our addresses based on the address width, which we could do with something like `format_args!("{:#0width$x}", self.0, width = mem::size_of::<usize>() * 2)`.

hal-core/src/addr.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    /// Returns `true` if `self` is aligned on the specified alignment.
+    fn is_aligned<A: Into<usize>>(self, _align: A) -> bool {
+        unimplemented!("eliza")
+    }
+
+    fn as_ptr(&self) -> *const () {
+        unimplemented!("eliza")
+    }
```
mystor: Hmm, given that we'll be running with memory virtualisation enabled, it may not make sense to support converting a physical address to a raw pointer, as derefing it will never do what you want.

hal-core/src/addr.rs
hawkw: i changed it to return `*const T` in my paging branch:
https://github.com/hawkw/mycelium/blob/93188e8e171e99c3a4f991e3fdfecdefcc0ecc8a/hal-core/src/lib.rs#L81-L84

so i didn't want to touch it here. 

i'm also open to changing it to return `*mut T`; my thought process was that it was better to default to the slightly more restrictive `*const`, and if you want mutability, you have to cast to `*mut` explicitly (or we add an `as_mut_ptr` method). this may not matter.

hal-core/src/addr.rs
hawkw: maybe. i didn't want to make unrelated code changes in this branch beyond moving stuff around.

hal-core/src/addr.rs
hawkw: i'm fine with changing the fmt impl, but i just snarfed it out of the hal_x86_64 crate and pasted it here...would prefer to make any behavioral or implementation changes separately.

hal-core/src/addr.rs
hawkw: i think this is from a `s/u64/usize`

hal-core/src/addr.rs
mystor: Most types in the stdlib which deal with raw pointers just work with `*mut T`, as by the point where you're working with raw pointers, the extra syntactic salt isn't really helpful, and can obscure the actual dangerous code. 

e.g. `NonNull` has the variance properties of `*const T`, but the `as_ptr` method always returns a `*mut T`. https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.as_ptr

hawkw:
I've made this PR no longer a draft. I think the feedback from @mystor on address types should probably be addressed separately.

src/lib.rs
```diff
@@ -10,7 +10,10 @@ use hal_core::{boot::BootInfo, mem};
 
 use alloc::vec::Vec;
 
-pub fn kernel_main(bootinfo: &impl BootInfo) -> ! {
+mod wasm;
+
+const HELLOWORLD_WASM: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/helloworld.wasm"));
+
```
mystor: I'm pretty sure this change won't build, unless I'm missing something else in this interdiff.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -125,7 +68,7 @@ pub fn init(bootinfo: &impl hal_core::boot::BootInfo<Arch = crate::X64>) -> &'st
     tracing::info!("intializing IDT...");
 
     unsafe {
-        IDT.register_handlers::<TestHandlersImpl>().unwrap();
+        IDT.register_handlers::<H>().unwrap();
```
iximeow: buried lede, this is good

hawkw:
"whoop's"

src/lib.rs
hawkw: uwu, i did a widdle fucky wucky

hawkw: should be fixed in 9b85256c2deb0b1a6cd6bfd18233f57306527542

hawkw:
@iximeow ci is worky now

hawkw:
unless @mystor or @iximeow have any remaining blockers, i'm gonna go ahead and merge this

PR #61, opened 2020-01-16
decode x86 instructions around where we fault

iximeow:
it decodes instructions! and prints them!

~we've talked about putting this in the oops fn, which i think would be much better, but by the time we're in `oops()` the only context we have is some `Display` thing. and even there, the cause we get is a call site is a `core::panic::Location`, which gets us a line and column of the offending source, but not an actual instruction pointer :(~

this punts decoding and displaying to the architecture-specific `oops` fn, i tried stitching it into `Context` in some form but trying to have a mutable iterator-shaped thing available from a `Context` is beyond what i have patience to get into the type system.

we might be able to do something with `backtrace` to get instruction pointers, and that probably lets us get function start addresses so we could actually disassemble some instructions before the fault address, like @mystor was hoping - hopefully the faulting instruction wasn't reached from a conditional!

also yaxpeax is crates now so i can just add 33 lines and it works and that owns

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -405,6 +405,8 @@ dependencies = [
  "tracing",
  "wasmi",
  "wat",
+ "yaxpeax-arch",
+ "yaxpeax-x86",
 ]
 
 [[package]]
@@ -1002,3 +1004,22 @@ name = "winapi-x86_64-pc-windows-gnu"
 version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
+
+[[package]]
+name = "yaxpeax-arch"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "effe4b97c0b314b9fa460f2513cd3eb5ffaf9643cc13c3307476b84f92e54175"
+dependencies = [
+ "num-traits",
+]
+
+[[package]]
+name = "yaxpeax-x86"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b8cf2aeb6c9b75d7e4ada5b26bcfa3ca7102b6b7c70a6a1ea2bcc59738b7229"
+dependencies = [
+ "num-traits",
+ "yaxpeax-arch",
+]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,6 +38,8 @@ mycotest = { path = "mycotest" }
 
 [target.'cfg(target_arch = "x86_64")'.dependencies]
 hal-x86_64 = { path = "hal-x86_64" }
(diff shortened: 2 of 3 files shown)


hawkw:
yeah, that's why `oops` as written takes a `dyn Display`...perhaps in the future it will take multiple ones that get stitched together in a preset way, idk.

iximeow:
> and implements Display by disassembling the instructions around that instruction pointer,

i tried doing this and eventually gave up - if we decide "the `Display` impl is ten instructions" it might be plausible, but that feels a bit arbitrary to commit to like that. anything more general would want to punt display and processing to whatever's actually doing displaying, which means exposing types (or `dyn Display`). i had issues trying to fit an instruction type into `Context`, which seems kind of silly anyway, so i'm not sold on that being a great direction to go. and `dyn` gets weird because we probably don't want to like, `Box<dyn Iterator<Item=Box<dyn Display>>>` when handling a world-ending panic..

anyway for the time being i just put the logic in `oops` which seems plenty for a first-brush "what went wrong". if we want to think about processing instructions (f.ex symbolicating or walking the stack or ... ) we'll want to extend some interfaces later.

also our deps broke because asm is `llvm_asm!` now, not `asm!`

hawkw:
i'm guessing this has probably bitrotted, but it would still be nice to have if you wanna get it working?

iximeow:
wow this pr can get a birthday cake soon

hawkw:
> wow this pr can get a birthday cake soon

@katie-martin-fastly

katie-martin-fastly:
> > wow this pr can get a birthday cake soon
> 
> @katie-martin-fastly

:cake: :cake: :cake:

iximeow:
so i updated this to use yaxpeax-x86 1.0, but i don't know how to get the kernel to fault to see it work again. so uh. if you wanna merge it as-is, i _think_ it works but idk how to test it does

hawkw:
> so i updated this to use yaxpeax-x86 1.0, but i don't know how to get the kernel to fault to see it work again. so uh. if you wanna merge it as-is, i _think_ it works but idk how to test it does

oh whoa i missed this message >.<

hawkw:
> so i updated this to use yaxpeax-x86 1.0, but i don't know how to get the kernel to fault to see it work again.

one good way to make the kernel fault is making it overflow its stack :) :) :) :)
```rust
#[inline(never)]
fn totally_normal_and_safe_fine_function_that_doesnt_do_anything_bad() {
    totally_normal_and_safe_fine_function_that_doesnt_do_anything_bad()
}
```
or something

hawkw:
im just gonna fuckin merge this assuming it still compiley

hawkw (approved):
im just gonna fuckin merge this assuming it still compiley!!!!

we can't really test anything that writes to the vga buffer right now since the bootloader update in #104 broke the vga buffer code and i haven't written framebuffer code to replace it. so i'm just gonna assume this works lol

hawkw:
lmao this PR was open for 582 days. Big Mycelium Energy.

PR #63, opened 2020-01-20
style: change x86 inline asm from AT&T syntax to intel

hawkw:
As discussed in [#59 (comment)][1], it's good to standardize on a
syntax, and @mystor and I both have a minor preference for Intel over
AT&T syntax. Also, when the work started in PR #61 merges, we will be
showing disassembly in crash screens using Intel syntax, so we should
use consistent syntax in source.

This PR changes inline assembly to use Intel syntax. In cases where the
inline assembly is a single instruction with no syntactic differences
between AT&T and Intel syntax (e.g. "cli", "hlt", et cetera), I did not
add the "intel" flag to the `asm!` macro. I can if we'd prefer this to
be explicit, but the syntax is currently equivalent.

[1]: https://github.com/hawkw/mycelium/pull/59#discussion_r366014379


hawkw:
asm_syntax

hawkw:
@iximeow any hot takes?

PR #65, opened 2020-02-07
feat(testing): add basic section-based test declaration support

mystor:
This is an outline of a potential section-based approach for declaring tests which supports running tests from across the dependency graph with a single test runner.

May supercede #59

Example Output:
```
 [INFO] run tests
 [INFO] | test, test.name="basic_alloc", test.module="mycelium_kernel"
 [INFO] | |- mycelium_kernel: vec=[], vec.addr=0x8
 [INFO] | |- mycelium_kernel: vec=[5], vec.addr=0x3fc058
 [INFO] | |- mycelium_kernel: vec=[5, 10], vec.addr=0x3fc048
 [INFO] | test, test.name="wasm_hello_world", test.module="mycelium_kernel"
[TRACE] | | invoke_index, index=0, args=RuntimeArgs([I32(1), I32(0), I32(1), I32(20)])
 [INFO] | | | fd_write, fd=1, iovs=0, iovs_len=1, nwritten=20
 [INFO] | | | |- mycelium_kernel::wasm::wasi: fd_write, buf_slice=[104, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100, 10], buf_str=Ok("hello world\n")
 [INFO] | test, test.name="it_works", test.module="mycelium_util::testing"
 [INFO] | |- mycelium_util::testing: I'm running in a test!
 [WARN] |- mycelium_kernel: 3 passed | 0 failed
```


util/src/testing.rs
```diff
@@ -0,0 +1,95 @@
...
+// FIXME: Declare a `#[test]` custom attribute instead?
+#[macro_export]
+macro_rules! decl_test {
+    (fn $name:ident $($t:tt)*) => {
+        fn $name $($t)*
+
+        // Introduce an anonymous const to avoid name conflicts. The `#[used]`
+        // will prevent the symbol from being dropped, and `link_section` will
+        // make it visible.
+        const _: () = {
```
mystor: Unfortunately, we can't put this behind a `#[cfg(test)]`, as that cfg flag won't be visible in dependencies. We may need to use `RUSTFLAGS="--cfg mycelium_test"` (or something like that) to turn on test code for the whole dependency tree.

util/src/testing.rs
```diff
@@ -0,0 +1,95 @@
...
+// instances.
+extern "C" {
+    static __start_MyceliumTests: ();
+    static __stop_MyceliumTests: ();
+}
+
+/// Get a list of test objects.
+pub fn all_tests() -> &'static [Test] {
+    unsafe {
+        // FIXME: These should probably be `&raw const __start_*`.
```
hawkw: is `&raw` new proposed syntax? i would love a link to this as I've never seen it before!

util/src/testing.rs
```diff
@@ -0,0 +1,95 @@
+use core::fmt;
+use core::mem;
+use core::slice;
+
+/// Internal definition type used for a test.
+pub struct Test {
```
hawkw: nit/tioli: since this has all pub fields, should there maybe be a private `_p: ()` or something to stop it from being constructed from anywhere?

util/src/testing.rs
```diff
@@ -0,0 +1,95 @@
+use core::fmt;
+use core::mem;
+use core::slice;
+
+/// Internal definition type used for a test.
```
hawkw: nit: the test runner in the kernel consumes this, so i am not sure if i would call it "internal" :P

util/src/testing.rs
hawkw: that makes sense, i would really like to avoid linking these into non-test builds eventually. this is fine for now...

util/src/testing.rs
```diff
@@ -0,0 +1,95 @@
...
+pub struct Test {
+    pub module: &'static str,
+    pub name: &'static str,
+    pub run: fn() -> bool,
```
hawkw: nit: WDYT about making `run` private, and adding a

```rust
impl Test {
    pub fn run(&self) -> bool {
        (self.run)()
    }
}
```
so that code consuming this can treat it as a method?

src/lib.rs
```diff
@@ -103,3 +127,40 @@ pub static GLOBAL: mycelium_alloc::Alloc = mycelium_alloc::Alloc;
 fn alloc_error(layout: core::alloc::Layout) -> ! {
     panic!("alloc error: {:?}", layout);
 }
+
+#[cfg(target_os = "none")]
+#[panic_handler]
+#[cold]
+fn panic(panic: &core::panic::PanicInfo) -> ! {
```
hawkw: i'm fine with moving this out of main, just curious if that was actually necessary or if you just did it because it made more sense?

src/arch/x86_64.rs
```diff
@@ -159,3 +159,22 @@ pub fn oops(cause: &dyn core::fmt::Display) -> ! {
...
+/// Exit using `isa-debug-exit`, for use in tests.
+///
+/// NOTE: This is a temporary mechanism until we get proper shutdown implemented.
+#[cfg(test)]
+pub(crate) fn qemu_exit(exit_code: QemuExitCode) {
+    let code = exit_code as u32;
+    unsafe {
+        asm!("out 0xf4, eax" :: "{eax}"(code) :: "intel","volatile");
+    }
+}
```
hawkw: can we stick a 
```rust
#[cfg(test)]
qemu_exit(QemuExitCode::Failed);
```
at the end of `arch::x86_64::oops`, before the loop? that way, if a test panics/faults, we can report failure rather than hanging. ideally, we would be able to track which test failed, so that panics can be a nice way to fail tests rather than presenting less information, but it would be nice if panics/faults didn't make the CI build just _hang_

.github/workflows/rust.yml
```diff
@@ -22,10 +22,27 @@ jobs:
...
+      with:
+        profile: minimal
+        toolchain: nightly
+        components: rust-src, llvm-tools-preview
+    - name: install dev env
+      uses: actions-rs/cargo@v1.0.1
+      with:
+        command: dev-env
+    - name: install qemu
+      run: sudo apt-get update && sudo apt-get install qemu
```
hawkw: it would be nice if we could figure out a way to use the github actions cache thingy so we don't have to install qemu over and over...not a blocket though!

.github/workflows/rust.yml
```diff
@@ -22,10 +22,27 @@ jobs:
...
+    - name: run tests
```
hawkw: would be good to have a step that runs any "regular" (i.e. not in qemu) rust tests in crates as well. i think we define some now as of #64, so it would be good for them to run on CI as well. 

if this was a separate job, it could run in parallel with the qemu tests.

util/src/testing.rs
mystor: Yup! It's a way to do field indexing without requiring the object you're taking a reference to to be valid memory. (rfc: https://github.com/rust-lang/rfcs/pull/2582, tracking issue: https://github.com/rust-lang/rust/issues/64490)

util/src/testing.rs
mystor: It has to be constructed from within the macro, which is why it has all-pub fields, so I can't add a `_p: ()` or the `decl_test!` macro won't work.

It's technically harmless to construct it anywhere you like, it won't do anything unless it's in the correct section.

util/src/testing.rs
mystor: We can't do that, because then we wouldn't be able to construct this struct from within the `decl_test!` macro.

src/lib.rs
mystor: It's necessary because when the `[lib]` target is being built as a `harness = false` test target, it needs to have a panic handler defined, and the `main.rs` file isn't being included.

I tried making the `[[bin]]` target have the test (and turning off tests for the `[lib]` target), but that didn't work, as I couldn't `#[cfg(test)]` within `kernel_main`, as it was a separate crate.

src/arch/x86_64.rs
mystor: Yes, I was intending to do that, but forgot to do it before pushing. I'll add that.

.github/workflows/rust.yml
mystor: I didn't even know that was a thing, so I'm not sure I can help with that :-P
If you have some documentation for how the cache thing works to point me to, let me know.

.github/workflows/rust.yml
mystor: Yeah, we should probably do that. We'll probably want to make sure to exclude the root `mycelium_kernel` crate, though.

util/src/testing.rs
hawkw: ah, nvm, i missed that! seems like it could have a const-fn ctor but that might just be boilerplate?

src/lib.rs
hawkw: ah, that makes sense! i was just curious

.github/workflows/rust.yml
hawkw: we might have to just `cargo test -p $FOO` for every other crate...which is fine, it's just a little flaky since it needs to be manually updated as we add more crates...

.github/workflows/rust.yml
hawkw: we can figure this out later!

.github/workflows/rust.yml
mystor: I ended up just using more `#[cfg(...)]` attributes to stub out a dummy `fn main()` for the `mycelium_kernel` crate when we're running on the host platform.

util/src/testing.rs
mystor: Yeah, I think that might just end up being boilerplate. I'm inclined to just leave it as-is.

hawkw:
clippy seems broken for some reason and i don't know why? but otherwise, this is awesome

PR #66, opened 2020-02-08
feat(clippy-ci): fix clippy running on ci

mystor:
The `actions-rs/toolchain` action runs `rustup show` before installing the toolchain requested. Unfortunately, when `rustup show` is run, `rustup` will read the `rust-toolchain` file from the root of the project, and immediately download and install the toolchain mentioned there.

This means that the `nightly` toolchain is installed with no components specified, and can select a version which does not currently have the `clippy` component. As `rustup` will refuse to downgrade your rust nightly version, this leads to an incompatible version being installed, and tests failing.

This branch moves the toolchain installing step before the `actions/checkout` step, meaning that `rustup` can't see the `rust-toolchain` file, and doesn't install the incorrect toolchain version.


mystor:
Pushed some changes to fix the remaining warnings, so we should have a relatively clean log now.

alloc/src/lib.rs
```diff
@@ -31,6 +31,10 @@ static FREE: AtomicUsize = AtomicUsize::new(HEAP_SIZE);
 /// NOTABLE INVARIANTS:
 ///  * `Layout` is non-zero sized (enforced by `GlobalAlloc`)
 ///  * `align` is a power of two (enforced by `Layout::from_size_align`)
```
hawkw: nit/tioli: should these be moved to the "Safety" heading?

hal-x86_64/src/serial.rs
```diff
@@ -64,6 +64,7 @@ pub struct Nonblocking {
 }
 
 impl Port {
+    #[allow(clippy::unreadable_literal)]
```
hawkw: nit/TIOLI: this shoud probably just have an underscore stuck in it to placate clippy, rather than allowing the lint?

hal-x86_64/src/serial.rs
```diff
@@ -331,11 +332,6 @@ impl<'a> LockInner<'a> {
         self.inner.is_write_ready()
     }
 
-    #[inline(always)]
-    fn is_read_ready(&self) -> bool {
-        self.inner.is_read_ready()
-    }
```
hawkw: if this is unused i think that's a bug...there's probably someplace where we call `is_write_ready` when we mean to call `is_read_ready`?

util/src/io/cursor.rs
```diff
@@ -36,10 +39,7 @@ impl<T> Cursor<T> {
     /// is not empty. So writing to cursor starts with overwriting `Vec`
     /// content, not with appending to it.
     pub fn new(inner: T) -> Cursor<T> {
-        Cursor {
-            pos: 0,
-            inner: inner,
-        }
+        Cursor { pos: 0, inner }
```
hawkw: it's cool when clippy wants us to rewrite code that was vendored from libstd :P

hal-x86_64/src/lib.rs
```diff
@@ -6,13 +6,14 @@
 // memory, and these types don't implement Copy to ensure they aren't
 // inadvertantly copied.
 #![allow(clippy::trivially_copy_pass_by_ref)]
+#![allow(clippy::cognitive_complexity)]
```
hawkw: nit/tioli: it would be nice if places where we globally allow lints had a comment explaining why we need to allow that lint. not a blocker...

hal-x86_64/src/serial.rs
hawkw: oh, nvm, we do use it where we're supposed to, it's just that we call the method on `inner`:
https://github.com/hawkw/mycelium/blob/9403955dea67c43b6f7ce28ac270a51aae2df334/hal-x86_64/src/serial.rs#L201
we never use the forwarded impl on the wrapper, so removing this is fine 👍

hal-x86_64/src/serial.rs
mystor: Sure. I don't have strong opinions, but vaguely preferred the no-underscore version of the number, so stuck with it. I'll switch to adding the underscore.

PR #69, opened 2020-02-09
feat(ci): cache cargo-xbuild and bootimage binaries when running ci

mystor:
This is a fairly flawed system, and currently won't ever actually rebuild the generated binaries, but may be acceptable. It appears to improve the performance of CI builds by about 2 minutes, from 5 minutes to 3 minutes.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -13,10 +13,18 @@ jobs:
         toolchain: nightly
         components: rust-src, llvm-tools-preview
     - uses: actions/checkout@v2
+    - uses: actions/cache@v1
+      id: toolcache
+      with:
+        path: _devtools
+        key: devtools-${{ runner.os }}-v0
     - name: install dev env
+      if: steps.toolcache.outputs.cache-hit != 'true'
       uses: actions-rs/cargo@v1.0.1
       with:
         command: dev-env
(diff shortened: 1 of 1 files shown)


mystor:
Another potential option, for someone more familiar with docker, would be to have a pre-configured docker container with rust, qemu, and the tools we need installed, and do a `rustup update && cargo test-x64` inside it. I think there's a chance that if we have a pre-built one, it'll run faster. I have no idea how to make one, or if it'd even be faster, though.

mystor:
re-requesting review because running tests again didn't catch anything on fire.

.github/workflows/rust.yml
```diff
@@ -13,10 +13,18 @@ jobs:
         toolchain: nightly
         components: rust-src, llvm-tools-preview
     - uses: actions/checkout@v2
+    - uses: actions/cache@v1
+      id: toolcache
+      with:
+        path: _devtools
+        key: devtools-${{ runner.os }}-v0
```
mystor: I made sure that the cache key had a version # in it, so that we'd be able to change the key we're using & thus force re-fetching the cached binaries. We'd need to update the `v0` at the end of this key to a `v1`, and we'd re-build the tools we're using.

PR #71, opened 2020-03-28
feat(x86_64): initial paging implementation

hawkw:
This PR adds a very basic initial implementation of a paging subsystem
for x86_64. We can now pass some very simple tests for translating an
identity mapped address to itself, and remapping a page. There is a
bunch of stuff that will need to be implemented in follow-up branches.

The `hal-x86_64` paging module uses recursive mapping, and currently
relies on the 511th PML4 entry being the recursive entry. This is 
*not* what the `bootloader` crate's "recursive_page_table" feature 
does...so we don't use that feature. Instead, we just instruct the 
bootloader to map the physical memory, and we set up our own recursive
mapping in entry 511. 

The advantages of hard-coding the recursive entry are:

* The implementation is much simpler
* The recursive address doesn't need to be stored and passed around, 
  which probably has at least a small perf benefit.
* 511 is all 1s in binary, and is octal 0o777, so it is _really_ 
  obvious when printing ddresses in octal (and pretty obvious when 
  printing them in hex, too!)

However, this might not be great in the future if we want to work with
other bootloaders, etc., so we can revisit it in the future. It works
for now.

There is a *lot* of work remaining, but I would really prefer to merge
this PR and implement missing functionality in follow-ups. This is a
pretty huge chunk of code already and I'd like to get it merged back.

Signed-off-by: Eliza Weisman <elzia@buoyant.io>
Co-authored-by: Nika Layzell <nika@thelayzells.com>


hal-x86_64/src/mm/mod.rs
```diff
@@ -66,26 +68,33 @@ pub fn init_paging(vm_offset: VAddr, recursive_addr: VAddr) {
...
-        // if entry.is_present() {
-        tracing::trace!(idx, ?entry);
-        //     present_entries += 1;
-        // }
-    }
-    tracing::info!(present_entries);
+    // Create the recursive pagetable entry.
+    assert!(
+        !pml4.entries[RECURSIVE_INDEX].is_present(),
+        "bootloader must not have used entry 511"
```
hawkw, line 75: should we just clobber it if it did?

mystor: I don't think so, as that page entry inserted by the bootloader could be for the stack or program code, which would be very unfortunate to replace with a recursive pagetable entry.
I think it's better to crash than to risk that.

hawkw: yeah, we'd have to remap that entry somewhere else first.  eventually i _would_ rather not be panicking at boot-time due to (possibly non-deterministic?) bootloader behavior, but it doesn't _seem_ to ever want to use 511 in qemu tests, so, for now, who cares!

hal-core/src/boot.rs
```diff
@@ -21,6 +21,10 @@ pub trait BootInfo {
         None
     }
 
+    fn init_paging(&self);
+    // XXX(eliza): is this generic enough???
+    // fn phys_mem_offset(&self) -> crate::VAddr;
```
mystor, line 26: Can we remove this commented-out method?

hal-core/src/interrupt/mod.rs
```diff
@@ -5,6 +5,8 @@ pub use self::ctx::Context;
 
 /// An interrupt controller for a platform.
 pub trait Control {
+    type Registers: fmt::Debug + fmt::Display;
```
mystor, line 8: Isn't this from the other patch?

hal-core/src/mem/page.rs
```diff
@@ -53,6 +55,68 @@ pub unsafe trait Alloc<S: Size> {
...
+
+pub trait PageFlags<S: Size> {
+    fn set_writable(&mut self, writable: bool) -> &mut Self;
+    fn set_executable(&mut self, executable: bool) -> &mut Self;
+
+    // fn set_present(&mut self, present: bool) -> &mut Self;
+
+    fn is_writable(&self) -> bool;
+    fn is_executable(&self) -> bool;
+    // fn is_present(&self) -> bool;
```
mystor, line 115: Why is this commented out?

hal-x86_64/src/control_regs.rs
```diff
@@ -0,0 +1,32 @@
...
+        unsafe {
+            asm!("mov %cr3, $0" : "=r"(val));
+        };
+        let addr = PAddr::from_u64(val);
+        let pml4_page =
+            Page::starting_at(addr).expect("PML4 physical addr not aligned! this is very bad");
+        (pml4_page, Flags(val))
+    }
+
+    pub unsafe fn write(pml4: Page<PAddr, Size4Kb>, flags: Flags) {
```
mystor, line 19: clippy is unhappy about no documentation

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -140,50 +141,68 @@ impl hal_core::interrupt::Control for Idt {
...
+        ) {
+            unsafe {
+                // Safety: who cares!
+
+                crate::vga::writer().force_unlock();
+                if let Some(com1) = crate::serial::com1() {
+                    com1.force_unlock();
+                }
+            }
+            tracing::error!(code = ?&format_args!("{:x}", code), "lmao, a general protection fault is happening");
```
mystor, line 196: lol

Do we want to remove this before landing?

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+/// This value should only be set once, early in the kernel boot process before
+/// we have access to multiple cores. So, technically, it could be a `static
+/// mut`. But, using an atomic is safe, even though it's not strictly necessary.
+static VM_OFFSET: AtomicUsize = AtomicUsize::new(core::usize::MAX);
+
+impl PageTable<level::Pml4> {
+    fn current(vm_offset: VAddr) -> &'static mut Self {
+        let (phys, _) = crate::control_regs::cr3::read();
+        unsafe { Self::from_pml4_page(vm_offset, phys) }
+    }
```
mystor, line 104: It occurs to me that this should probably be `unsafe`, because there's no way to guarantee exclusivity, and you could pass in an arbitrary `vm_offset` value.

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+        &mut *(virt.as_ptr::<Self>() as *mut _)
+    }
+}
+
+impl<'mapper, A> Map<'mapper, Size4Kb, A> for PageCtrl
+where
+    A: page::Alloc<Size4Kb>,
+{
+    type Handle = Handle<'mapper, level::Pt>;
+    // type Handle =
```
mystor, line 120: nit: remove this comment

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+    /// Map the virtual memory page represented by `virt` to the physical page
+    /// represented bt `phys`.
+    ///
+    /// # Panics
+    ///
+    /// - If the physical address is invalid.
+    /// - If the page is already mapped.
+    fn map_page(
```
mystor, line 128: These methods should probably be `unsafe`.

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+impl PageCtrl {
+    pub fn current() -> Self {
+        let vm_offset = VM_OFFSET.load(Ordering::Acquire);
+        assert_ne!(
+            vm_offset,
+            core::usize::MAX,
+            "`init_paging` must be called before calling `PageTable::current`!"
+        );
+        let vm_offset = VAddr::from_usize(vm_offset);
+        let pml4 = PageTable::current(vm_offset);
```
mystor, line 193: We can probably replace this with a hardcoded value, but let's perhaps do this in a follow-up.

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+
+        Ok(())
+    }
+}
+
+mycelium_util::decl_test! {
+    fn identity_mapped_pages_are_reasonable() -> Result<(), ()> {
+        let ctrl = PageCtrl::current();
+
+        let page = VirtPage::<Size4Kb>::containing(VAddr::from_usize(0xb8000));
```
mystor, line 823: Let's maybe have a comment mentioning that the VGA page is identity-mapped.

hal-x86_64/src/serial.rs
```diff
@@ -122,6 +122,10 @@ impl Port {
             _is_blocking: PhantomData,
         }
     }
+
+    pub unsafe fn force_unlock(&self) {
```
mystor, line 126: I think this change is from the other patch, can we rebase on the new master?

hal-x86_64/src/mm/mod.rs
```diff
@@ -0,0 +1,831 @@
...
+
+    fn table_addr(v: VAddr) -> VAddr;
+}
+
+impl<L: Level> fmt::Debug for PageTable<L> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.debug_struct("PageTable")
+            .field("level", &format_args!("{}", L::NAME))
+            .field("addr", &format_args!("{:p}", self))
+            // .field("entries", &self.entries)
```
mystor, line 647: Can we drop this comment?

It might also be cool to have this eventually print all non-empty entries, but we don't need to do that yet.

hal-core/src/lib.rs
```diff
@@ -1,5 +1,7 @@
 #![cfg_attr(target_os = "none", no_std)]
+#![feature(try_trait)]
```
mystor, line 2: What do we need this for?

hal-x86_64/src/mm/mod.rs
hawkw: i was unsure about this one given the "creating a possiblly-invalid pointer is safe, dereferencing it is unsafe" heuristic but yeah.

hal-x86_64/src/serial.rs
hawkw: whoops, yeah, i forgot to rebase, thanks for catching

hal-x86_64/src/mm/mod.rs
hawkw: Hmm, I'm not sure. I think maybe the trait should probably be unsafe to implement, but I think that if the implementation upholds the required invariants, mapping a page is safe?

mystor: Mapping a page seems unsafe to me. You can do things like aliasing the stack onto the heap, making memory executable, and make read-only memory read-write.

Unmapping a page also seems unsafe to me, because combined with mapping a page, you can break tons of invariants expected by the language.

hal-core/src/lib.rs
hawkw: I no longer remember. I suspect that means "we don't".

hal-x86_64/src/mm/mod.rs
hawkw: Some of these seem more like an argument for making the APIs for setting _flags_ on an entry unsafe (since that's how you would, e.g. make things writable or whatever)

hal-core/src/lib.rs
hawkw: ah, it's needed so that we can refer to `core::option::NoneError` here:
https://github.com/hawkw/mycelium/blob/87881769d12019c64e51c69e9b3c8489866f47e4/hal-core/src/mem/page.rs#L372-L376

which...i think we can just remove.

hal-x86_64/src/mm/mod.rs
hawkw: Yeah, locking is probably a good call.

hal-multiboot2/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "hal-multiboot2"
+version = "0.1.0"
+authors = ["Eliza Weisman <eliza@buoyant.io>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
```
hawkw, line 9: ASDF ghghghhghgh aaaaah why are you back???

PR #72, opened 2020-04-18
fix(x86_64): forcibly unlock COM1 when oopsing

hawkw:
Currently, the kernel oops handler will forcibly unlock the VGA buffer
before writing the oops message. Ths prevents deadlocks if an oops
occurs while it was locked. However, we also write traces to COM1, but
we don't unlock it when oopsing.

This commit fixes this, by ensuring we forcibly unlock both VGA and
COM1 when oopsing. I also refactored the x64 ISRs a bit so that the
interrupt context for faults is printed by the oops function, rather than
by the ISR. This means we don't have to worry about force-unlocking
in multiple places.

### DIFF
--- a/hal-x86_64/src/interrupt/pic.rs
+++ b/hal-x86_64/src/interrupt/pic.rs
@@ -48,9 +48,10 @@ impl CascadedPic {
 }
 
 impl hal_core::interrupt::Control for CascadedPic {
+    type Registers = super::Registers;
     fn register_handlers<H>(&mut self) -> Result<(), hal_core::interrupt::RegistrationError>
     where
-        H: Handlers,
+        H: Handlers<super::Registers>,
     {
         Err(RegistrationError::other(
             "x86_64 handlers must be registered via the IDT, not to the PIC interrupt component",
--- a/src/arch/x86_64.rs
+++ b/src/arch/x86_64.rs
@@ -2,7 +2,7 @@ use bootloader::bootinfo;
 use core::sync::atomic::{AtomicUsize, Ordering};
 use hal_core::{boot::BootInfo, mem, PAddr};
 pub use hal_x86_64::{interrupt, NAME};
-use hal_x86_64::{serial, vga};
+use hal_x86_64::{interrupt::Registers as X64Registers, serial, vga};
 
 #[derive(Debug)]
 pub struct RustbootBootInfo {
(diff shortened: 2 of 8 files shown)


src/arch/x86_64.rs
```diff
@@ -90,6 +114,8 @@ impl hal_core::interrupt::Handlers for InterruptHandlers {
     where
         C: hal_core::interrupt::ctx::Context + hal_core::interrupt::ctx::CodeFault,
     {
+        // Safety: we are crashing. Don't worry about it.
+        unsafe { force_unlock_writers() }
         tracing::error!(registers = ?cx.registers(), "double fault",);
         oops(&format_args!("  DOUBLE FAULT\n\n{}", cx.registers()))
```
mystor, line 120: I feel like we should probably do the `force_unlock_writers` calls within `oops`, and not log anything directly from within these handlers. Perhaps we should add an `interrupt: Option<&dyn Context>` argument to `oops`?

src/arch/x86_64.rs
```diff
@@ -130,18 +156,14 @@ pub extern "C" fn _start(info: &'static bootinfo::BootInfo) -> ! {
 #[cold]
 pub fn oops(cause: &dyn core::fmt::Display) -> ! {
     use core::fmt::Write;
+    // Safety: we are oopsing. Don't worry about it.
+    // We may unlock writers twice, if we oopsed from an ISR, but...who cares.
+    unsafe { force_unlock_writers() }
 
+    tracing::error!(%cause, "oopsing");
     unsafe { asm!("cli" :::: "volatile") }
```
mystor, line 164: Should we `cli` before we unlock the writers, to avoid an external interrupt after we unlock?

src/arch/x86_64.rs
hawkw: > Perhaps we should add an `interrupt: Option<&dyn Context>` argument to `oops`?

slapping myself for not thinking about that, god.

hawkw: @mystor changed it to work that way in 2f213a5d4e87a4e40c138a228994402a677bef0c. Ended up needing a somewhat larger change to ensure that all the `Registers` associated types lined up.

hal-x86_64/src/interrupt/pic.rs
```diff
@@ -48,9 +48,10 @@ impl CascadedPic {
 }
 
 impl hal_core::interrupt::Control for CascadedPic {
+    type Registers = super::Registers;
     fn register_handlers<H>(&mut self) -> Result<(), hal_core::interrupt::RegistrationError>
     where
-        H: Handlers,
+        H: Handlers<super::Registers>,
```
hawkw, line 54: Kinda hate that this forces the PICs to pretend to be aware of CPU exception stack frames...but I think the whole interrupt handling interface that currently exists will need a bigger refactor before it makes sense for paripheral-driven interrupts (PIC should probably implement a different trait...)

src/arch/x86_64.rs
```diff
@@ -66,32 +66,37 @@ impl BootInfo for RustbootBootInfo {
...
-
 pub(crate) struct InterruptHandlers;
 
-impl hal_core::interrupt::Handlers for InterruptHandlers {
+/// Forcibly unlock the IOs we write to in an oops (VGA buffer and COM1 serial
+/// port) to prevent deadlocks if the oops occured while either was locked.
+///
+/// # Safety
+///
+///  /!\ only call this when oopsing!!! /!\
```
mystor, line 76: I think this comment is in the wrong place now.

hal-x86_64/src/interrupt/pic.rs
mystor: That seems like a reasonable plan to me!

PR #76, opened 2020-05-02
feat(util): add exponential backoffs to spin loops

hawkw:
This PR adds exponential backoffs to spin loops. When backing off, we
will issue an exponentially increasing number of `spin_loop_hint`s
(`pause` instructions on x86), up to 2^8 times per spin.

This may improve power consumption while spinning.

### DIFF
--- a/util/src/sync/mod.rs
+++ b/util/src/sync/mod.rs
@@ -19,13 +19,16 @@ impl Backoff {
     pub(crate) const DEFAULT_MAX_EXPONENT: u8 = 8;
 
     pub(crate) const fn new() -> Self {
-        Self { exp: 0, max: 8 }
+        Self {
+            exp: 0,
+            max: Self::DEFAULT_MAX_EXPONENT,
+        }
     }
 
     /// Returns a new exponential backoff with the provided max exponent.
     #[allow(dead_code)]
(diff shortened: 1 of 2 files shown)


util/src/sync/mod.rs
```diff
@@ -7,3 +7,44 @@ pub use loom::sync::atomic;
...
+pub(crate) struct Backoff {
+    exp: u8,
+    max: u8,
+}
+
+impl Backoff {
+    pub(crate) const DEFAULT_MAX_EXPONENT: u8 = 8;
+
+    pub(crate) const fn new() -> Self {
+        Self { exp: 0, max: 8 }
```
mystor, line 22: Should we use `DEFAULT_MAX_EXPONENT` here for `max:`?

util/src/sync/spin.rs
```diff
@@ -33,9 +33,10 @@ impl<T> Mutex<T> {
     }
 
     pub fn lock(&self) -> MutexGuard<'_, T> {
+        let mut boff = super::Backoff::default();
         while self.locked.compare_and_swap(false, true, Ordering::Acquire) {
             while self.locked.load(Ordering::Relaxed) {
-                spin_loop_hint();
```
mystor, line 38: Clippy wants you to remove the now-unused `spin_loop_hint` import.

PR #78, opened 2020-05-06
feat(hal): allow representing dynamically sized pages

hawkw:
Currently, the `Size` trait only allows representing statically known
sizes. This interface makes sense when _mapping_ pages, since the users
of that API will likely be saying things like "I want to map this 4kb
virtual page to a 4kb physical page". However, it makes less sense when
_translating_ pages/addresses. If I ask the page table "what page is
this address on?", I currently have to know in advance the size of the
page, which doesn't make sense — that's part of the information I would
be trying to _query_ by asking for the page containing an address. The
current trait does not model this nicely.

This commit updates the `Size` trait in `hal_core::mem::Page` to allow
representing dynamically sized pages, with a `Size::size()` fn rather
than an associated constant. The previous constant-only `Size` trait is
now a new trait, `StaticSize`, which _implements_ `Size`; this allows
code which requires a statically known size to bound type parameters
with `StaticSize` rather than `Size`.

Unfortunately, `size` fields on structs can no longer be `PhantomData`,
as they may in some cases be...real data; however, the current set of
types implementing `StaticSize` are still all zero-sized so this
shouldn't matter _that_ much.

### DIFF
--- a/hal-core/src/mem/page.rs
+++ b/hal-core/src/mem/page.rs
@@ -1,10 +1,17 @@
 use crate::{Address, PAddr, VAddr};
-use core::{cmp, fmt, marker::PhantomData, ops, slice};
+use core::{cmp, fmt, ops, slice};
 
-pub trait Size: Copy + Eq + PartialEq {
+pub trait Size: Copy + Eq + PartialEq + fmt::Display {
+    /// The size (in bits) of this page.
+    fn size(&self) -> usize;
+}
+
+/// A statically known page size.
+pub trait StaticSize: Copy + Eq + PartialEq + fmt::Display {
     /// The size (in bits) of this page.
     const SIZE: usize;
     const PRETTY_NAME: &'static str;
+    const INSTANCE: Self;
 }
 
 pub type TranslateResult<A, S> = Result<Page<A, S>, TranslateError<S>>;
@@ -113,7 +120,7 @@ where
     /// same address.
     fn identity_map(&'mapper mut self, phys: Page<PAddr, S>, frame_alloc: &mut A) -> Self::Handle {
         let base_paddr = phys.base_address().as_usize();
-        let virt = Page::containing(VAddr::from_usize(base_paddr));
+        let virt = Page::containing(VAddr::from_usize(base_paddr), phys.size());
         unsafe { self.map_page(virt, phys, frame_alloc) }
     }
 }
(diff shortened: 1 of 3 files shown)


hawkw:
if anyone has better names for any of the things in this patch, i would welcome your suggestions

hawkw:
(i think i wanted this for frame alloc reasons but no longer remember why; we definitely need something like this before we do anything serious w/ paging though)

hawkw:
@iximeow do you care about this or shall i merge it?

hawkw:
> good change imo too, what are the serious paging use cases that might want variable page sizes though?

looking up what page a virtual address is on is the obvious one — we only know the size of the page once we walk the page table to find the page. requiring the OS to always annotate address translations would mean requiring that all OSes only ever use one page size.

in general, i suspect we will _mostly_ hand out 1GB pages, but i can see a few use-cases where we might want to allocate smaller pages:
 - wasms that advertise that they will use less than 1GB of memory in their metadata. of course, such a wasm might _lie_, but in that case, we can defer allocating additional pages for when it actually asks us to (or kill it, if this is considered a misbehavior that deserves the death penalty?)
 - memory mapped IO regions that need to be identity mapped but are fairly small might be mapped with less granularity
 - if the kernel doesn't need a lot of memory to represent the kernel-side data for a process, we might want smaller pages? idk.

there may be other reasons an OS might occasionally care about smaller pages when the hardware supports huge pages, idk. maybe look @ what lunix does?

hawkw:
> good change imo too, what are the serious paging use cases that might want variable page sizes though?

looking up what page a virtual address is on is the obvious one — we only know the size of the page once we walk the page table to find the page. requiring the OS to always annotate address translations would mean requiring that all OSes only ever use one page size.

in general, i suspect we will _mostly_ hand out 1GB pages, but i can see a few use-cases where we might want to allocate smaller pages:
 - wasms that advertise that they will use less than 1GB of memory in their metadata. of course, such a wasm might _lie_, but in that case, we can defer allocating additional pages for when it actually asks us to (or kill it, if this is considered a misbehavior that deserves the death penalty?)
 - memory mapped IO regions that need to be identity mapped but are fairly small might be mapped with less granularity
 - if the kernel doesn't need a lot of memory to represent the kernel-side data for a process, we might want smaller pages? idk.

there may be other reasons an OS might occasionally care about smaller pages when the hardware supports huge pages, idk. maybe look @ what lunix does?

PR #81, opened 2020-09-05
page range, page, and mem region API tweaks & fixes

hawkw:
This fixes some annoying stuff in the mem APIs, makes end addrs
inclusive as expected, and implements a couple missing APIs. Factored
this out of [my page allocator PR][1], which will eventually be rebased
onto this.

[1]: https://github.com/hawkw/mycelium/tree/eliza/page-alloc

### DIFF
--- a/hal-core/src/mem/mod.rs
+++ b/hal-core/src/mem/mod.rs
@@ -64,12 +64,14 @@ impl<A: Address> Region<A> {
     /// Returns `true` if `self` contains the specified address.
     pub fn contains(&self, addr: impl Into<A>) -> bool {
         let addr = addr.into();
-        addr > self.base && addr < self.end_addr()
+        addr >= self.base && addr < self.end_addr()
     }
 
-    /// Returns the last address (inclusive) of the memory region.
+    /// Returns the end address (exclusive) of the memory region.
+    ///
+    /// This is the start address of the next memory region.
     pub fn end_addr(&self) -> A {
-        self.base + (self.size - 1)
+        self.base + self.size
     }
 
     pub fn kind(&self) -> RegionKind {
@@ -81,7 +83,7 @@ impl<A: Address> Region<A> {
         size: S,
     ) -> Result<page::PageRange<A, S>, page::NotAligned<S>> {
         let start = page::Page::starting_at(self.base, size)?;
-        let end = page::Page::starting_at(self.end_addr() + 1, size)?;
+        let end = page::Page::starting_at(self.end_addr(), size)?;
         Ok(start.range_to(end))
     }
 
--- a/hal-core/src/mem/mod.rs
+++ b/hal-core/src/mem/mod.rs
@@ -67,9 +67,9 @@ impl<A: Address> Region<A> {
         addr > self.base && addr < self.end_addr()
     }
 
-    /// Returns the end address of the memory region.
+    /// Returns the last address (inclusive) of the memory region.
     pub fn end_addr(&self) -> A {
-        self.base + self.size
+        self.base + (self.size - 1)
     }
 
     pub fn kind(&self) -> RegionKind {
--- a/hal-core/src/mem/mod.rs
+++ b/hal-core/src/mem/mod.rs
@@ -76,8 +76,22 @@ impl<A: Address> Region<A> {
         self.kind
     }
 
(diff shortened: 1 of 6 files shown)


hal-core/src/mem/mod.rs
```diff
@@ -67,17 +67,31 @@ impl<A: Address> Region<A> {
         addr > self.base && addr < self.end_addr()
     }
 
-    /// Returns the end address of the memory region.
+    /// Returns the last address (inclusive) of the memory region.
     pub fn end_addr(&self) -> A {
-        self.base + self.size
+        self.base + (self.size - 1)
     }
```
mystor, line 73: Why make this inclusive? one-past-the-end seems like a more common and convenient "end" value to be working with

hal-core/src/mem/page.rs
```diff
@@ -2,13 +2,13 @@ use crate::{Address, PAddr, VAddr};
 use core::{cmp, fmt, ops, slice};
 
 pub trait Size: Copy + Eq + PartialEq + fmt::Display {
-    /// The size (in bits) of this page.
-    fn size(&self) -> usize;
+    /// Returns the size (in bytes) of this page.
+    fn in_bytes(&self) -> usize;
```
mystor, line 6: I think it is more common to word this as `as_bytes` or `to_bytes`, given it takes `&self`, and returns another interpretation of the value.

That being said, I suppose that name might suggest that it's returning the size as `&[u8]`, which would be a bit weird. Not a huge deal either way.

Perhaps `as_usize` would be reasonable here (if less descriptive)?

hal-core/src/mem/page.rs
```diff
@@ -249,7 +252,7 @@ impl<A: Address, S: Size> Page<A, S> {
 
     pub fn contains(&self, addr: impl Into<A>) -> bool {
         let addr = addr.into();
-        addr >= self.base && addr <= self.end_address()
+        addr >= self.base && addr <= self.end_addr()
     }
```
mystor, line 256: if you make `end_addr` be non-inclusive again, this should be changed to be `addr < self.end_addr()`.

hal-x86_64/src/interrupt/mod.rs
```diff
@@ -55,7 +55,7 @@ pub fn init<H: Handlers<Registers>>() -> Control {
     tracing::info!("configuring 8259 PIC interrupts...");
 
     unsafe {
-        PIC.set_irq_addresses(0x20, 0x28);
+        PIC.set_irq_addres(0x20, 0x28);
```
mystor, line 58: why `addres`? It seems like you're trying to normalize to `addr` in most other places.

hawkw: this is a find-and-replace accident (replacing "address" with "addr" turns addresses" into "addres"). thanks for noticing that!

hal-core/src/mem/mod.rs
hawkw: i could go either way — honestly the goal was just to make the `contains` check not be broken, but we could just change it there. to me, inclusive feels more intuitive, but i could be convinced otherwise. @iximeow  wanna bikeshed this?

hal-core/src/mem/page.rs
hawkw: my thought process is that numbers that represent a value with a unit ought to be returned by methods with that unit in their names, and that, as you said, `as_bytes`/`to_bytes` really sound like conversions to a byte slice/array/whatever. the guidelines for conversions don't really deal with _unit_ conversions, only type conversions, so i picked `in_` kinda arbitrarily. but, i also liked that you get a nice little sentence --- `page.size().in_bytes()` and similar. 

`as_usize` seems like strictly less information, to me. but i'm open to change it if you don't like this.

hawkw:
@iximeow do you have opinions on the bikesheds?

hal-core/src/mem/mod.rs
hawkw: changed everything to inclusive in 7687749

hal-core/src/mem/page.rs
hawkw: changed it to `as_usize`

hawkw:
@mystor I think I've made all the changes you suggested, anything else that needs to be fixed or can I go ahead and merge this?

PR #82, opened 2020-09-05
feat(build): migrate from cargo xbuild to build-std feature

mystor:
I think this should fix the issues we were having with CI.


hawkw (approved):
yayyyy thank you 🙌

PR #83, opened 2020-09-05
feat(ci): make clippy happy again

mystor:


hawkw (approved):
this is probably "style:" and not "feat:" but lgtm otherwise

mystor:
Ah, makes sense. I'm not super familiar with this pr titling scheme, so was largely trying to mimic

hawkw:
the idea is that it could be used to generate changelogs, but in practice, none of this software has ever been "released", so who cares?

PR #86, opened 2020-12-19
feat(alloc): buddy block page frame allocator that kinda sorta works

hawkw:
this branch adds a working page frame allocator based on the buddy-block
allocation algorithm. it seems to more or less work, and can allocate
both single pages and ranges of pages.

however, it doesn't currently ensure correct alignments for page sizes
larger than the architecture's minimum page size --- there's no
difference, currently, between allocating one 2MB page and allocating a
2MB range of 4KB pages. for the 2MB page to be usable, though, it would
have to be aligned on a 2MB boundary, while the 2MB range of 4KB pages
only needs to start on a 4K boundary.

i'd like to add that in a future PR. for now, it just returns an error
when allocating page sizes that would require larger alignments, which
is fine for now i guess.

there's probably additional cleanup we could do, but whatever, i'm tired
of working on this and i want to merge it. we can make it better later.

a huge thank you to @mystor for help catching a couple of really
silly mistakes i made and getting this over the finish line!

Co-authored-by: Nika Layzell <nika@thelayzells.com>


src/lib.rs
```diff
@@ -48,6 +57,21 @@ pub fn kernel_main(bootinfo: &impl BootInfo) -> ! {
...
+                    // // XXX(eliza): this sucks; the page allocator should be less
+                    // // stupid.
+                    // let aligned_end = region.end_addr().align_down(4096usize);
+                    // let aligned_size = region.base_addr().difference(aligned_end) as usize;
+                    // let lost = region.size() - aligned_size;
+                    // if lost > 0 {
+                    //     tracing::warn!("leaking {} bytes to preserve alignment", lost);
+                    // }
+                    // let region =
+                    //     mem::Region::new(region.base_addr(), aligned_size, mem::RegionKind::FREE);
```
hawkw, line 71:
```suggestion
```

hawkw (commented):
MARRY FUCKIN CHRISTMAS

hawkw:
LOL THE GODDAMN COMPILER UPDATED WHILE I WAS WORKING ON THIS AND THE FEATURE FLAGS I ADDED ARE NO LONGER NECESSARY LOL I LOVE NIGHTLY RUST, THIS IS FINE

hawkw:
huh, tests passed for me locally, wtf???

hawkw:
okay now it works?

util/src/math.rs
```diff
@@ -0,0 +1,25 @@
+pub trait Log2 {
+    /// Returns the log base 2 of `self`.
+    fn log2(self) -> Self;
+}
+
+impl Log2 for usize {
+    /// Fast log base 2 implementation for integers.
```
mystor, line 7: Perhaps we should make sure to say this is log2, rounding up?

FWIW I think log2 rounding down can be a bit more efficient (though also a bit more gross, at least in my first pass, and only if it works for our use-case here) due to there basically being an instruction for that. For example an implementation like this one comes out to only a couple of instructions:

```rust
pub const fn log2_floor(u: usize) -> usize {
  const WIDTH: usize = mem::size_of::<usize>() * 8;
  WIDTH - (u | 1).leading_zeros() as usize - 1
}
```

```asm
example::log2_floor:
        or      rdi, 1
        bsr     rax, rdi
        ret
```
https://godbolt.org/z/r6oWqK

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+}
+
+// ==== impl BuddyAlloc ===
+
+impl BuddyAlloc {
+    #[cfg(not(loom))]
+    pub const fn new_default(min_size: usize) -> Self {
+        Self::new(
+            min_size,
+            // haha this is cool and fun
```
mystor, line 57: haha

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+use mycelium_util::trace;
+
+#[derive(Debug)]
+pub struct BuddyAlloc<L = [spin::Mutex<List<Free>>; 32]> {
+    /// Minimum allocateable page size in bytes.
+    ///
+    /// Free blocks on free_lists[0] are one page of this size each. For each
+    /// index higher in the array of free lists, the blocks on that free list
+    /// are 2x as large.
+    pub min_size: usize,
```
mystor, line 25: Why is this field `pub`? Feels like it probably shouldn't be given that mutating it would invalidate e.g. `min_size_log2`.

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+            vm_offset: AtomicUsize::new(0),
+            min_size_log2: mycelium_util::math::usize_const_log2(min_size),
+            heap_size: AtomicUsize::new(0),
+            free_lists,
+        }
+    }
+
+    pub fn set_vm_offset(&self, offset: VAddr) {
+        self.vm_offset
+            .compare_exchange(0, offset.as_usize(), AcqRel, Acquire)
```
mystor, line 111: This can be a `.swap` too, but given it's happening once during startup, I don't think it really matters...

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+    // TODO(eliza): nicer way to configure this?
+    fn offset(&self) -> usize {
+        let addr = self.vm_offset.load(Relaxed);
+        debug_assert_ne!(addr, 0, "you didn't initialize the heap yet dipshit");
+        addr
+    }
+
+    /// Returns the size of the allocation for a given order
+    fn size_for_order(&self, order: usize) -> usize {
+        1 << (self.min_size_log2 as usize + order)
```
mystor, line 125: isn't `self.min_size_log2` already `usize`?

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+    }
+
+    /// Returns the actual size of the block that must be allocated for an
+    /// allocation of `len` pages of `page_size`.
+    fn size_for<S: Size>(&self, page_size: S, len: usize) -> Result<usize> {
+        let mut size = page_size.as_usize() * len;
```
mystor, line 131: Should this be a checked mul?

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+            region = Region::new(new_base, region.size(), RegionKind::FREE);
+        }
+
+        // Is the size of the region a power of two? The buddy block algorithm
+        // requires each free block to be a power of two.
+        if !size.is_power_of_two() {
+            // If the region is not a power of two, split it down to the nearest
+            // power of two.
+            let prev_power_of_two = prev_power_of_two(size);
+            tracing::debug!(prev_power_of_two, "not a power of two!");
```
mystor, line 218: Won't `prev_power_of_two` be a power of two? It might be more useful to log out `size` here.

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+            buddy.offset = trace::hex(buddy_offset),
+            buddy.addr = ?buddy,
+        );
+
+        if core::ptr::eq(buddy as *const _, block.as_ptr() as *const _) {
+            tracing::trace!("buddy block is the same as self");
+            return None;
+        }
+
+        if unsafe { (*buddy).magic == Free::MAGIC } {
```
mystor, line 319: Is this sound? What's stopping us from accidentally putting the free magic value into the right place in the block while it's in use?

hal-core/src/mem/page/alloc.rs
```diff
@@ -0,0 +1,571 @@
...
+                // buddy may want to merge with it. When the buddy tries to
+                // remove the block from the free list, though, it will be
+                // surprised to find that the block is not, in fact, free!
+                //
+                // Therefore, we need to clobber the block magic, even if it
+                // will *probably* be immediately overwritten by whoever
+                // allocated the block --- they *might* not overwrite it, or
+                // someone else might free the buddy of this block before the
+                // user of this block writes to it!
+                block.make_busy();
```
mystor, line 418: This is scary that we're depending on the user not using the free magic in their heap-allocated payload D:

mystor (commented):
I haven't gone over this as thoroughly as usual, but I tried to leave some comments. It generally seems reasonable, but I don't really know a ton about how buddy-block allocators work off the top of my head, so I wouldn't have noticed algorithmic bugs in the implementation.

hal-core/src/addr.rs
```diff
@@ -100,7 +100,7 @@ pub trait Address:
 
     /// Returns `true` if `self` is aligned on the specified alignment.
     fn is_aligned<A: Into<usize>>(self, align: A) -> bool {
-        self.align_down(align) == self
+        self.as_usize() % align.into() == 0
```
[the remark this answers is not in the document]
hawkw: yeah, that's probably a good call. we probably should have a debug assertion that the alignment is a power of two, though, since this accepts an arbitrary number as an alignment...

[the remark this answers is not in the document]
hawkw: i think in the version that checks against the align of a type we can probably rely on rustc not making types with weird alignment

util/src/math.rs
hawkw: yeah, the buddy block algorithm relies on this being `ceil(log2(x))`; this actually caused a bug when i used a fast impl that rounded down. i'll rename the function.

hal-core/src/mem/page/alloc.rs
hawkw: we already logged the size of the region a few lines ago:
https://github.com/hawkw/mycelium/blob/aaaf737e1fd92d11a5d0fa01df0b2d4e3c70555d/hal-core/src/mem/page/alloc.rs#L202

hal-core/src/mem/page/alloc.rs
hawkw: It was intended for debug stats etc--- I thought it was okay because the allocator is immutable when it's in use (it locks the free lists internally). I should probably just make it an accessor though.

hal-core/src/mem/page/alloc.rs
hawkw: So, technically, the free list is the authoritative source for whether or not a block is free, the magic is just a hint so that we don't have to traverse the whole free list when a block is definitely busy. However, we currently panic if the magic indicates that the block is free, but it's not on the free list...which is incorrect way to handle a false positive. 

I changed it in 14ec7b9dc343ed728d9d74081dd2488f61e36a37 so that the free magic is *strictly* an optimization hint, rather than relying on it for correctness.

PR #87, opened 2020-12-20
feat(util): add `InitOnce` and `Lazy` cells in `sync`

hawkw:
This branch adds new `InitOnce<T>` and `Lazy<T>` cells in
`mycelium_util::sync`, to model statics that must be initialized at
runtime. `InitOnce` is manually initialized to a value, while `Lazy` is
constructed with an initializer. For particularly performance-sensitive
use cases, `InitOnce` also offers an unsafe `get_unchecked` method,
which assumes that the cell has already been uninitialized, and elides
the checks. This is, of course, quite unsafe.


util/src/sync/init_once.rs
```diff
@@ -0,0 +1,296 @@
...
+
+    #[track_caller]
+    #[inline]
+    fn deref(&self) -> &Self::Target {
+        match self.get() {
+            Some(t) => t,
+            None => panic!("InitOnce<{}> not yet initialized!", any::type_name::<T>(),),
+        }
+    }
+}
```
mystor, line 160: I don't love using a panicking `Deref` impl, as calls to `Deref` can be inserted without writing anything in the source text. Perhaps we could have `.get()` with the panicking behaviour, and `.try_get()` which returns an `Option` (or perhaps `unwrap`/`get`?)

util/src/sync/init_once.rs
```diff
@@ -0,0 +1,296 @@
...
+                    INITIALIZED,
+                    Ordering::AcqRel,
+                    Ordering::Acquire,
+                ) {
+                    unreachable!(
+                        "Lazy<{}>: state changed while locked. This is a bug! (state={})",
+                        any::type_name::<T>(),
+                        actual
+                    );
+                }
```
mystor, line 234: Oh, why use `compare_exchange` here? Seems like because it's a bug for it to have changed, we should just `.store(INITIALIZED, Ordering::AcqRel)`, or `.swap` if we want to inspect the old value for assertions.

util/src/sync/init_once.rs
```diff
@@ -0,0 +1,296 @@
...
+            }
+            Err(state) => unreachable!(
+                "Lazy<{}>: unexpected state {}!. This is a bug!",
+                any::type_name::<T>(),
+                state
+            ),
```
mystor, line 240: I suppose if we want to we could make this a `unreachable_unchecked` in release builds. Perhaps we should include a `debug_unreachable_unchecked!()` (better name pending) in the `util` library to do an `unreachable!` which is actually unchecked in release builds?

util/src/sync/init_once.rs
hawkw: yeah this could be swap, good point!

util/src/sync/init_once.rs
hawkw: I like _having_ a `Deref`, though...it's much more of a pain to use these without one. But I see your point...

hawkw: I like *having* a deref impl, though...otherwise, it makes these types much more of a PITA to use.  But, you're probably right that it's not a great idea.

PR #89, opened 2020-12-21
chore: don't use `add-path` for caching tools on CI

hawkw:
GitHub Actions has [deprecated the `add-path` command][1] which we are
currently using for caching `cargo install` binaries. This means our CI
builds are now failing.

This branch updates CI to use `actions-rs/install` for installing and
caching `cargo install` binaries instead. Hopefully this fixes CI.

Additionally, I've used YAML anchros to reduce repetition in the CI
configs.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -75,8 +75,8 @@ jobs:
       uses: actions-rs/cargo@v1.0.1
       with:
         command: check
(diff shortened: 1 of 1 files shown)


mystor:
oh nice! either I didn't realize `actions-rs/install` had caching support or it didn't when I implemented the CI setup before. This is much cleaner.

PR #94, opened 2021-01-24
fix buddy block merging

hawkw:


hal-core/src/mem/page/alloc.rs
```diff
@@ -482,10 +517,13 @@ where
...
             if let Some(mut buddy) = unsafe { self.take_buddy(block, curr_order, &mut free_list) } {
                 // Okay, merge the blocks, and try the next order!
+                if buddy < block {
+                    core::mem::swap(&mut block, &mut buddy);
+                }
                 unsafe {
                     block.as_mut().merge(buddy.as_mut());
                 }
-                tracing::trace!("merged with buddy");
+                tracing::trace!("merged with buddy {:?} {:?}", block, buddy);
```
hawkw, line 526: nit: prefer structured fields:
```suggestion
                tracing::trace!(?block, ?buddy, "merged with buddy");
```

hawkw (commented):
thank you so much for catching this bug! i left a few nits

hawkw:
@mystor also it looks like rustfmt needs to be run?

PR #95, opened 2021-01-25
feat(alloc): impl `GlobalAlloc` for `buddy::Alloc`

hawkw:
this PR adds an impl of `GlobalAlloc` for `mycelium_alloc::buddy::Alloc`
so that we can reuse the buddy allocator as a heap allocator. no
functional change here.

in order to add this, i refactored the allocation code out of the
`page::Alloc` impl into inherent methods that operate on `Layout`s and
addrs, and reimplemented page allocation on top of that.

since we now operate on `Layout`s internally, we also rely on some of
their additional invariants.

note that i haven't actually tested this, but, like...it probably works?

### DIFF
--- a/alloc/src/buddy.rs
+++ b/alloc/src/buddy.rs
@@ -459,7 +459,11 @@ where
     L: AsRef<[spin::Mutex<List<Free>>]>,
     S: Size + core::fmt::Display,
 {
-    /// Allocate a range of `len` pages.
+    /// Allocate a range of at least `len` pages.
+    ///
+    /// If `len` is not a power of two, the length is rounded up to the next
+    /// power of two. The returned `PageRange` struct stores the actual length
+    /// of the allocated page range.
     ///
     /// # Returns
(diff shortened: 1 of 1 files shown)


hawkw:
lol welp i guess we actually _do_ need to do extra preprocessing of page range allocation requests -> `Layout`s, because i actually wrote a test that tries to allocate a page range with a weird length on purpose to see how badly it dies. good work, past eliza!

do we actually care about being able to allocate numbers of pages that would do a misaligned allocation (like, for example, "ten")? should we handle this by silently aligning the page count up to a good alignment, or should it be an error? note that since the page allocation methods return page ranges with metadata, the user can look at the actual number of pages that was allocated and go "oh, it's more than i asked for". meanwhile, the page allocation errors just totally throw out everything<sup>1</sup>. cc @mystor wdyt?

---
<sup>1</sup>: we should probably add messages to them i guess

hawkw:
i'm gonna say we probably want to uh align up

PR #97, opened 2021-03-28
fix(util): make intrusive list links never Unpin

hawkw:
In light of rust-lang/rust#82834, we must ensure that the intrusive
linked list pointers never get mutable-noalias optimizations (see also
rust-lang/rust#63818). Adding a `PhantomPinned` to the `Links` struct
ensures it will always be `!Unpin`, disabling mutable-noalias.


iximeow (approved):
learning cool facts about the rust language today, wow

mystor (approved):
Looks good!

PR #100, opened 2021-03-28
chore(ci): ensure the right qemu bin is installed

hawkw:
Apparently `apt-get install qemu` *used* to provide the
 `qemu-system-x86_64` binary, but maybe it doesn't anymore? Maybe
explicitly trying to install that package will fix CI?


mystor (approved):
woo, looks like it fixed it :-)

PR #101, opened 2021-05-16
feat(hal): add methods to map/unmap page ranges

hawkw:
This PR adds methods to the `hal_core::mem::page::Map` trait to map,
unmap, and identity map ranges of pages at a time. These methods have
default implementations.

This will make my life easier when I start integrating the `acpi` crate.

Note that these actually *don't* have to be atomic, since the `Map`
trait takes `&mut self`, ensuring that the pages won't be mapped by
someone else, and we handle all page mapping errors by panicking, so we
don't need to e.g. roll back changes if there's a failure.

Since the default impl requires mutably borrowing the mapper in a loop,
I changed the `Map` trait a bit so that it no longer has a generic
lifetime. The actual borrowing of the page table entry is handled by a
`Handle` struct, now, and the entry associated type just has to tell it
how to mutate the entry's flags. This should make implementing the
mapper trait much easier since there isn't a bunch of restrictive
lifetime nonsense.


hawkw:
@iximeow do you wanna take a look at this and tell me how hilariously broken it is, or should i just merge it without writing any tests?

PR #104, opened 2021-06-06
feat: use the new version of bootloader

hawkw:
this updates mycelium to the new version of `bootloader` (v0.10.x) which
we need so that the bootloader can give us the RDSP address so we can do
ACPI things easily. also it has cool uefi stuff in it.

unfortunately the new bootloader crate requires you to basically write
your own bespoke build tool, WHICH I HAVE DONE WITH EXCESSIVE VIOLENCE.
now we have our own terrible build tool, which i have named `inoculate`,
because [thats how u make mushrooms][1].

inoculate can now sort of build a mycelium disk image, run mycelium in QEMU,
and run tests. also, i added a cool little program for running the tests and capturing
output from each test separately, and determining which test failed. it can even do
a fake version of `cargo test`'s output: 

image

also i had to fix a bunch of stuff that's broken with the new bootloader. the
vga buffer nops now because the bootloader switches to the framebuffer...

also @mystor's wasm hello world test is busted now, it page faults.
gotta figure out why but it would be nice to merge this i guess?

my head hurts a lot.

[1]: https://fungi.com/blogs/articles/a-simplified-overview-of-mushroom-cultivation-strategies


.cargo/config
```diff
@@ -1,10 +1,10 @@
...
-run-x64 = "run -Z build-std=core,alloc --target=x86_64-mycelium.json"
-debug-x64 = "run -Z build-std=core,alloc --target=x86_64-mycelium.json -- -gdb tcp::1234 -S"
-test-x64 = "test -Z build-std=core,alloc --target=x86_64-mycelium.json"
+run-x64 = "run -Z build-std=core,alloc --target=x86_64-mycelium.json -- run"
+debug-x64 = "run -Z build-std=core,alloc --target=x86_64-mycelium.json -- run -- -gdb tcp::1234 -S"
+test-x64 = "test -Z build-std=core,alloc --target=x86_64-mycelium.json -- test"
 clippy-x64 = "clippy -Z build-std=core,alloc --target=x86_64-mycelium.json"
-bootimage-x64 = "bootimage -Z build-std=core,alloc --target=x86_64-mycelium.json"
+build-x64 = "run -Z build-std=core,alloc --target=x86_64-mycelium.json"
+inoculate = "run -Z build-std=core,alloc --target=x86_64-mycelium.json"
```
mystor, line 10: github is angry that you don't have a newline at the end of this file :-P

Cargo.toml
```diff
@@ -26,9 +28,15 @@ hal-core = { path = "hal-core" }
 mycelium-alloc = { path = "alloc", features = ["buddy", "bump"] }
 mycelium-util = { path = "util" }
 rlibc = "1.0"
+# NOTE FOR FUTURE ELIZAS WHO ARE MESSING WITH THIS: the bootloader crate's build
+# script is not that good, and breaks if you put this in `cfg(...).dependencies`
+# instead of normal [dependencies]. don't move this.
+bootloader = { version = "0.10.6" }
```
mystor, line 34: oof

hal-core/Cargo.toml
```diff
@@ -8,4 +8,4 @@ edition = "2018"
 
 [dependencies]
 tracing = { version = "0.1", default_features = false }
-mycelium-util = { path = "../util" }
+mycelium-util = { path = "../util" }
```
mystor, line 11: more angry missing newlines

hal-x86_64/src/mm/mod.rs
```diff
@@ -287,19 +299,48 @@ impl<R: level::Recursive> PageTable<R> {
...
+        let _e = span.enter();
+        let entry = &mut self[idx];
+        tracing::trace!(?entry);
+        if !entry.is_present() {
+            tracing::debug!("entry not present!");
+            return None;
+        }
+        // XXX(eliza): should we have a different return type to distinguish
+        // between "the entry is huge, so you stop the page table walk" and "the entry is not
+        // present"? we definitely should...
```
mystor, line 324: yeah, that sounds like a solid idea :-)

inoculate/src/cargo.rs
```diff
@@ -0,0 +1,34 @@
...
+        return Ok(cargo);
+    }
+
+    let cargo = PathBuf::from(env!("CARGO"));
+    ensure!(
+        cargo.exists(),
+        "cargo path from build-time CARGO env var ({}) does not exist",
+        cargo.display()
+    );
+    Ok(cargo)
```
mystor, line 33: I'm surprised we need to record the build time cargo env var here. Does it not work if we use the `CARGO` env var?

inoculate/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+        // if !output.status.success() {
+        if !output.success() {
+            // let stderr = String::from_utf8_lossy(&output.stderr);
+            return Err(format_err!(
+                "bootloader's builder command exited with non-zero status code"
+            ))
+            // .with_section(move || stdout.trim().to_string().header("stdout:"))
+            // .with_section(move || stderr.trim().to_string().header("stderr:"))
+            .suggestion("if you had gotten all the inputs right, this should have worked");
+        }
```
mystor, line 156: I don't suppose we can drop the commented-out sections here?

src/arch/x86_64.rs
```diff
@@ -131,15 +139,27 @@ impl hal_core::interrupt::Handlers<X64Registers> for InterruptHandlers {
...
-    crate::kernel_main(&bootinfo);
+bootloader::entry_point!(arch_entry);
+
+pub fn arch_entry(info: &'static mut boot_info::BootInfo) -> ! {
+    if let Some(offset) = info.physical_memory_offset.into_option() {
+        // Safety: i hate everything
+        unsafe {
+            vga::init_with_offset(offset);
+        }
+    }
```
mystor, line 156: I'm not sure I understand why we need to force-unlock the VGA buffer here. Why would it be locked?

src/arch/x86_64.rs
```diff
@@ -190,14 +210,94 @@ pub fn oops(
     let _ = vga.write_str("\n  it will never be safe to turn off your computer.");
 
     #[cfg(test)]
-    qemu_exit(QemuExitCode::Failed);
+    {
+        if let Some(test) = ptr::NonNull::new(CURRENT_TEST.load(Ordering::Acquire)) {
+            if let Some(com1) = serial::com1() {
```
mystor, line 215: A bit of me wonders if we should have a second `-serial` command in the test runner which we use for test progress signalling etc, so we don't need to worry as much about parsing it out of main serial output.

Would make it harder to split up serial output to be grouped by which test failed, I suppose.

src/lib.rs
```diff
@@ -12,7 +12,7 @@ use core::fmt::Write;
 use hal_core::{boot::BootInfo, mem};
 use mycelium_alloc::buddy;
 
-mod wasm;
+// mod wasm;
```
mystor, line 15: :cry:

util/src/trace.rs
```diff
@@ -14,6 +14,7 @@ where
 ///
 /// # Examples
 /// ```
+/// panic!("lol");
```
mystor, line 17: lol

mystor (commented):
I'll be honest, my eyes glazed over for parts of this as I don't know the bootstrapping parts of mycelium that well, but it generally looks good! :-) - Will definitely need to look into why the wasm code is failing, as that's a bit confusing.

inoculate/src/cargo.rs
hawkw: afaict it is always set, but i figured it was worth having a fallback in case it somehow isn't — the `bootloader` crate's examples just use the build-time env var so it probably doesn't really matter?

src/arch/x86_64.rs
hawkw: i don't think it's necessary, this was just scattershot debugging

.cargo/config
hawkw, line 10:
```suggestion
inoculate = "run -Z build-std=core,alloc --target=x86_64-mycelium.json"

```

hal-core/Cargo.toml
hawkw, line 11:
```suggestion
mycelium-util = { path = "../util" }

```

src/arch/x86_64.rs
hawkw: i thought about that --- this feels easier to just get it working, but we can keep messing with it.

PR #106, opened 2021-06-12
feat(gdb): support running gdb directly from inoculate

mystor:


inoculate/src/qemu.rs
```diff
@@ -101,8 +101,48 @@ impl Cmd {
...
+        // If the `--gdb` flag was passed, try to run gdb & connect to the kernel.
+        if qemu_settings.gdb {
+            let mut gdb = Command::new("gdb");
+
+            // Set the file, and connect to the given remote, then advance to `kernel_main`.
+            gdb.arg("-ex").arg(format!("file {}", binary.display()));
+            gdb.arg("-ex")
+                .arg(format!("target remote :{}", qemu_settings.gdb_port));
+            gdb.arg("-ex").arg("tbreak mycelium_kernel::kernel_main");
+            gdb.arg("-ex").arg("continue");
```
hawkw, line 127: what does this flag do? could be nice to have a comment explaining it?

inoculate/src/lib.rs
```diff
@@ -94,7 +94,7 @@ impl Options {
         &self,
         bootloader_manifest: &Path,
         kernel_manifest: &Path,
-    ) -> Result<PathBuf> {
+    ) -> Result<(PathBuf, PathBuf)> {
```
hawkw, line 97: nit/tioli: maybe worth turning this into a struct with named fields because it's not super obvious what the tuple of two pathbufs means? but, it's only used in one place so i don't care about this that much --- we can fix it later.

hawkw (approved):
lgtm, some non-blocking nits

inoculate/src/qemu.rs
```diff
@@ -101,8 +101,48 @@ impl Cmd {
     }
 
     #[tracing::instrument(skip(self))]
-    pub fn run_qemu(&self, image: &Path) -> Result<()> {
-        // TODO(eliza): should we `which qemu` here?
+    fn spawn_qemu(&self, qemu: &mut Command, binary: &Path) -> Result<Child> {
+        let (Cmd::Run { qemu_settings, .. } | Cmd::Test { qemu_settings, .. }) = self;
```
hawkw, line 105: if we're running tests, we should probably add the `--no-shutdown` QEMU flag, so that it doesn't abruptly terminate and make gdb crash, right? (run mode adds that flag automagically)

hawkw (changes requested):

inoculate/src/lib.rs
```diff
@@ -90,19 +90,23 @@ impl Options {
             .suggestion("have you tried not having it be missing")
     }
 
+    pub fn wheres_the_kernel_bin(&self) -> Result<PathBuf> {
+        tracing::debug!("where's the kernel binary?");
+        self.kernel_bin
+            // the bootloader crate's build script gets mad if this is a
+            // relative pathe
```
hawkw, line 97: typo (yes, it's my typo and you just copied and pasted it, lol)
```suggestion
            // relative path
```

hawkw (approved):

PR #109, opened 2021-06-17
feat(util): add bit packing utilities

hawkw:
this PR adds some nifty utilities for packing bits into integers. 

unlike some other bit-packing code that exists currently, these are
const-eval compatible: most bit packing/unpacking operations can be
performed in `const fn`, and when bit packing is performed at runtime,
the specs for packing and unpacking a particular bit range can be
defined in `const`s. this means that whenever possible, bit masks can be
computed at compile time, and the packing/unpacking ops that occur at
runtime are *just* a bitand/bitor and a shift. initially wrote them for
my double-fault TSS branch, but thought they should merge in a separate
PR beforehand.

### DIFF
--- a/util/proptest-regressions/bits.txt
+++ b/util/proptest-regressions/bits.txt
@@ -1,0 +1,14 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
+#
+# It is recommended to check this file in to source control so that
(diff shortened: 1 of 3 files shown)


iximeow (approved):
bitvec-rs is undoubtedly sad but const eval is zero costier than "really fast but not quite const eval sometimes" so this is good

hawkw:
also like bitvec would be good if we ever had to pack bits into anything bigger than a word (i would never)

PR #112, opened 2021-06-27
feat(x86_64): use a separate IST stack for double faults

### LINKED ISSUE #56: x86: separate stack for double faults
so that we don't boot loop when the kernel overflows its stack (which might reasonably happen!), the CPU pushes an exception frame to call into the page fault handler, and the kernel overflows its stack again, causing a double fault, and the CPU pushes an exception frame to call into the double fault handler, and ... you get my drift.

we can do this using the TSS "interrupt stack table", setting an interrupt stack table offset in the IDT descriptor for the double-fault ISR, and having the IST entry at that offset point at an okay stack.

the page fault handler should maybe also have its own okay stack in the IST for similar reasons (and it's much more helpful to debug a page fault oops screen with info about the page fault than a "we did a double fault somehow lol" oops screen)

hawkw:
## motivation

so that we don't boot loop when the kernel overflows its stack (which
might reasonably happen!), the CPU pushes an exception frame to call
into the page fault handler, and the kernel overflows its stack again,
causing a double fault, and the CPU pushes an exception frame to call
into the double fault handler, and ... you get my drift.

we can do this using the TSS "interrupt stack table", setting an
interrupt stack table offset in the IDT descriptor for the double-fault
ISR, and having the IST entry at that offset point at an okay stack.

## solution

this branch implements the use of a separate interrupt stack for double
faults. in order to do this, it was necessary to add a bunch of new code
in `hal-x86_64` for configuring the GDT and segmentation registers.
previously, we just relied on the `bootloader` crate to set this stuff
up for us, but in order to use the TSS, we have to actually be able
to interact with these structures.

closes #56

## future work

as i discussed in issue #56, we may eventually also want to reserve a
separate IST stack for the page fault ISR. stack overflows will often
result in a page fault, and it would be nice to trap those into the
page fault handler, rather than letting the page fault handler also
overflow the stack and go straight to the double fault handler. that
way, we could print more useful debugging info for the page fault,
instead of just "lol we double faulted lmao".


hawkw:
lol this branch is over a year old lmao

PR #113, opened 2021-09-25
quick framebuffer driver

hawkw:
wow whoa it actually works!
image

### DIFF
--- a/hal-core/src/framebuffer.rs
+++ b/hal-core/src/framebuffer.rs
@@ -1,0 +1,155 @@
+use core::ops::{Deref, DerefMut};
+
+pub trait Draw {
+    /// Return the width of the framebuffer in pixels.
+    fn width(&self) -> usize;
+
+    /// Returns the height of the framebuffer in pixels.
+    fn height(&self) -> usize;
+
+    /// Set the pixel at position (`x`, `y`) to the provided `color`.
+    fn set_pixel(&mut self, x: usize, y: usize, color: RgbColor) -> &mut Self;
+
+    /// Draw a horizontal line of length `len` at height `y`.
+    ///
+    /// By default, this method calls `set_pixel` in a loop. This _works_, but
+    /// implementations can almost certainly provide a more optimal
+    /// implementation, and are thus encouraged to override this method.
+    fn line_horiz(&mut self, y: usize, len: usize, color: RgbColor) -> &mut Self {
+        for x in 0..len {
+            self.set_pixel(x, y, color);
+        }
+        self
+    }
+
+    /// Draw a vertical line of length `len` at column `y`.
+    ///
+    /// By default, this method calls `set_pixel` in a loop. This _works_, but
+    /// implementations can almost certainly provide a more optimal
+    /// implementation, and are thus encouraged to override this method.
+    fn line_vert(&mut self, x: usize, len: usize, color: RgbColor) -> &mut Self {
+        for y in 0..len {
+            self.set_pixel(x, y, color);
+        }
+        self
+    }
+
+    #[inline]
+    fn fill_row(&mut self, y: usize, color: RgbColor) -> &mut Self {
+        self.line_horiz(y, self.width(), color)
+    }
+
+    #[inline]
+    fn fill_col(&mut self, x: usize, color: RgbColor) -> &mut Self {
+        self.line_vert(x, self.height(), color)
+    }
+
+    /// Fill the entire framebuffer with the provided color.
+    ///
+    /// By default, this method calls `set_pixel` in a loop. This _works_, but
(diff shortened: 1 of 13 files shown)


mystor:
Despite #114 apparently targeting this branch for changes, it seems like many of the changes in this branch are also visible in that PR's diff, so I don't know if/what should be reviewed in this PR separately.

hawkw:
> Despite #114 apparently targeting this branch for changes, it seems like many of the changes in this branch are also visible in that PR's diff, so I don't know if/what should be reviewed in this PR separately.

the difference is that i turned some files into `mod.rs` in #114, so those files appear to be "added". i probably should have just thought ahead and made them `mod.rs` here...

hawkw:
> > Despite #114 apparently targeting this branch for changes, it seems like many of the changes in this branch are also visible in that PR's diff, so I don't know if/what should be reviewed in this PR separately.
> 
> the difference is that i turned some files into `mod.rs` in #114, so those files appear to be "added". i probably should have just thought ahead and made them `mod.rs` here...

@mystor opened  #116 to switch to Rust 2018 style modules so that we don't have these weird diffs in the future, WDYT?

hal-x86_64/src/framebuffer/embedded_graphics.rs
```diff
@@ -0,0 +1,33 @@
...
+//         self.config.bounding_box()
+//     }
+// }
+
+// // impl<C> DrawTarget for FrameBuffer
+// // where
+// //     C: PixelColor,
+// // {
+
+// // }
```
mystor, line 33: I presume you delete this in #114?

hal-x86_64/src/framebuffer.rs
```diff
@@ -0,0 +1,101 @@
+use core::ptr;
+use hal_core::framebuffer::{Draw, RgbColor};
+
+#[derive(Debug)]
+pub struct Framebuffer<'buf, B> {
+    buf: B,
+    cfg: &'buf Config,
+}
```
mystor, line 8: This might be doable in `hal-core`, but I suppose it doesn't really matter either way.

hal-core/src/framebuffer.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub const BLUE: Self = Self {
+        red: 0,
+        green: 0,
+        blue: 255,
+    };
+}
+
+impl<'lock, D> Draw for mycelium_util::sync::spin::MutexGuard<'lock, D>
+where
+    D: Draw,
```
mystor, line 102: This impl is technically not necessary anymore, right?

hal-x86_64/src/framebuffer.rs
```diff
@@ -0,0 +1,101 @@
...
+    Rgb,
+    Gray,
+}
+
+#[cfg(feature = "embedded-graphics-core")]
+#[doc(cfg(feature = "embedded-graphics-core"))]
+mod embedded_graphics;
+#[cfg(feature = "embedded-graphics-core")]
+#[doc(cfg(feature = "embedded-graphics-core"))]
+pub use self::embedded_graphics::*;
```
mystor, line 33: Perhaps this should be dropped, as I don't think this is where it's going to end up. I suppose it's deleted in #114 so it doesn't matter.

hal-x86_64/Cargo.toml
```diff
@@ -11,3 +11,4 @@ log = ["tracing/log"]
 hal-core = { path = "../hal-core" }
 mycelium-util = { path = "../util" }
 tracing = {version = "0.1", default_features = false, features = ["attributes"] }
+embedded-graphics-core = { version = "0.3", optional = true }
```
mystor, line 14: I don't think this is used in the current patch. Shouldn't this be a dependency of `hal-core`? I suppose this is deleted in #114 so it doesn't matter.

util/src/macros.rs
```diff
@@ -51,3 +51,21 @@ macro_rules! unreachable_unchecked {
...
+}
+
+/// Variadic version of [`core::cmp::min`].
+#[macro_export]
+macro_rules! min {
+    ($arg:expr) => { $arg };
+    ($arg1:expr, $($arg:expr),+) => {
+        core::cmp::min($arg1, $crate::min!( $($arg),+ ))
+    };
+}
```
mystor, line 71: I think these are unused now. Do we want to keep them around?

mystor (approved):

hal-x86_64/src/framebuffer.rs
hawkw: i also considered that, and may move it in the future

util/src/macros.rs
hawkw: imo they're nice to have but i can nuke them

PR #114, opened 2021-09-25
feat(kernel): draw text using `embedded-graphics`

hawkw:
Depends on #113

This branch adds support for text rendering directly to the framebuffer
using the `embedded-graphics` crate. We can also draw whatever other
arbitrary nonsense using that crate, but the important part is being
able to print things again. :)

image


hal-x86_64/src/framebuffer/mod.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.set_pixel_rgb(x, y, color)
+    }
+}
+
+impl PixelKind {
+    fn convert_rgb(self, RgbColor { red, green, blue }: RgbColor) -> [u8; 4] {
+        match self {
+            PixelKind::Bgr => [blue, green, red, 0],
+            PixelKind::Rgb => [red, green, blue, 0],
+            PixelKind::Gray => [mycelium_util::max!(red, green, blue), 0, 0, 0],
```
mystor, line 86: Flawless RGB -> Grayscale conversion :-)

It might be nice to do a proper rgb to luminosity conversion here with weighted averages, because it's not _too_ difficult to do on integers. I think something like this should work (and it looks like it only ends up using one `imul` which is nice https://godbolt.org/z/Whe6a9TfP):

```rust
pub fn rgb2l(r: u8, g: u8, b: u8) -> u8 {
  ((21 * (r as u32) + 72 * (g as u32) + 7 * (b as u32)) / 100) as u8
}
```

src/lib.rs
```diff
@@ -31,6 +31,69 @@ pub fn kernel_main(bootinfo: &impl BootInfo) -> ! {
...
+            small_style,
+            Alignment::Center,
+        )
+        .draw(&mut target)
+        .expect("never panics");
+
+        tracing::trace!("hahahaha yayyyy we drew a screen!");
+    }
+
     arch::interrupt::init::<arch::InterruptHandlers>();
```
mystor, line 97: I don't love that were setting up interrupts after doing test printing on the framebuffer. I feel like we should initialize it after interrupts, so that things like faults are caught properly.

src/arch/x86_64.rs
```diff
@@ -55,6 +66,37 @@ impl BootInfo for RustbootBootInfo {
...
 
+    fn framebuffer(&self) -> Option<Self::Framebuffer> {
+        let framebuffer = self.inner.framebuffer.as_ref()?;
+
+        static FRAMEBUFFER: InitOnce<spin::Mutex<Framebuffer>> = InitOnce::uninitialized();
+        Some(
+            FRAMEBUFFER
+                .get_or_else(move || {
+                    let info = framebuffer.info();
+                    let buf = framebuffer.buffer();
```
mystor, line 77: Given that we're mutating through this reference, shouldn't we use `buffer_mut`?

I suppose we don't own the actual `BootInfo` struct so we can't get a mutable reference to it here? Feels sketchy, especially because we're given a `&'static mut` reference to the `BootInfo` by the bootloader, but I suppose changing it would be tricky, and looking at the `BootInfo` struct, it seems like the `FrameBuffer` value is the only one which requires mutable access for meaningful operations.

Perhaps we should transfer the `FrameBuffer` out of the `BootInfo` while we have mutable access to it in `arch_entry` and pass ownership of the struct into our `Framebuffer` type, so we can use it directly (or I suppose pass a mutable reference into `kernel_main`)? I don't love doing more stuff before we've set up things like interrupt handlers though.

mystor (commented):
This is really cool!

hal-core/src/framebuffer/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+    fn height(&self) -> usize;
+
+    /// Set the pixel at position (`x`, `y`) to the provided `color`.
+    fn set_pixel(&mut self, x: usize, y: usize, color: RgbColor) -> &mut Self;
+
+    /// Draw a horizontal line of length `len` at height `y`.
+    ///
+    /// By default, this method calls `set_pixel` in a loop. This _works_, but
+    /// implementations can almost certainly provide a more optimal
+    /// implementation, and are thus encouraged to override this method.
```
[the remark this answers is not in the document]
hawkw: this is essentially a `// TODO(eliza): provide a more optimal implementation` :)

hawkw:
@mystor fwiw a lot of your review comments actually apply to the code added in #113, which this branch is based on, so i'll be addressing that feedback there :)

hal-core/src/framebuffer/mod.rs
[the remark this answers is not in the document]
mystor: :+1: :-)

src/arch/x86_64.rs
hawkw: @mystor is https://github.com/hawkw/mycelium/pull/113/commits/270c2351d304a44e2d6f734b179d3d533a7dd35f any less sketchy? it now involves a gross `transmute`, so i'm not sure if that's any better...

hawkw: hmm, but without the framebuffer we can't log anything or display an oops screen on fautls? so i think we should just "try really hard not to fault here"

src/lib.rs
hawkw: hmm, but without the framebuffer we can't log anything or display an oops screen on fautls? so i think we should just "try really hard not to fault here"

src/arch/x86_64.rs
hawkw: okay, https://github.com/hawkw/mycelium/pull/113/commits/82c1d9aafd68ee1ed9247a066fab3e744246896a should be _actually_ not sketchy. it required some weird refactoring and makes the interfaces somewhat stranger, but it actually does everything we want, and we _always_ get the `&'static mut [u8]` from the bootloader's framebuffer struct, and we don't have a weird transmute...

mystor (approved):

hal-core/src/framebuffer.rs
```diff
@@ -1,5 +1,12 @@
 use core::ops::{Deref, DerefMut};
 
+#[cfg(feature = "embedded-graphics-core")]
+#[doc(cfg(feature = "embedded-graphics-core"))]
+mod embedded_graphics;
+#[cfg(feature = "embedded-graphics-core")]
+#[doc(cfg(feature = "embedded-graphics-core"))]
+pub use self::embedded_graphics::*;
```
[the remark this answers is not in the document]
hawkw: i anticipated adding more stuff in the module later

PR #115, opened 2021-09-25
chore: update to nightly-2021-09-22, fix clippy

hawkw:
This just updates Rust to the latest nightly, and fixes a new Clippy
lint on that nightly.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2021-06-05"
+channel = "nightly-2021-09-22"
 components = [
     "clippy",
     "rustfmt",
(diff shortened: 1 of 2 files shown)


mystor:
As we discussed on Discord, this is likely due to some changes to how rustc handles the `#[used]` attribute since https://github.com/rust-lang/rust/pull/87570, and has also had an impact on `linkme`, which uses the same linker section tricks. Changing the linker and linker flavour to `ld` works around the issue for now, due to being a less-clever linker, but it would be nice to fix for `lld` as well.

It may be necessary for rust to expose some way to use the old `llvm.used` attribute rather than `llvm.compiler.used` to make this work long-term.

PR #117, opened 2021-09-26
chore: update nightly to 2021-09-22

hawkw:
so, per @mystor, the nightly bump works if we switch to linking with GNU `ld`.

i spent a while struggling with trying to get rust-lld to work by passing the `-z nostart-stop-gc` argument, but `rust-lld` claims it doesn't know about that.

after adding this to the target JSON:
```json
    "linker-flavor": "ld.lld",
    "linker": "rust-lld",
    "post-link-args": {
        "ld.lld": [
            "-z nostart-stop-gc"
        ]
    },
```
i get
```
 <...massive horrible linker invocation> "--gc-sections" "-z nostart-stop-gc"
  = note: rust-lld: error: unknown -z value:  nostart-stop-gc


error: could not compile `mycelium-kernel` due to previous error
```
which seems surprising given that
> Require -z start-stop-gc, which is available in trunk ld.lld (https://reviews.llvm.org/D96914) and GNU ld newer than 2021-03-01 (https://sourceware.org/bugzilla/show_bug.cgi?id=27451)
(per some random googling i did https://reviews.llvm.org/D96914)

so, whatever, GNU `ld` works. but it would be nice if we could make rust-lld work again...

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2021-06-05"
+channel = "nightly-2021-09-22"
 components = [
     "clippy",
     "rustfmt",
--- a/util/src/io/impls.rs
+++ b/util/src/io/impls.rs
@@ -258,7 +258,7 @@ impl Write for &mut [u8] {
     #[inline]
     fn write(&mut self, data: &[u8]) -> io::Result<usize> {
         let amt = cmp::min(data.len(), self.len());
-        let (a, b) = mem::replace(self, &mut []).split_at_mut(amt);
+        let (a, b) = mem::take(self).split_at_mut(amt);
         a.copy_from_slice(&data[..amt]);
         *self = b;
         Ok(amt)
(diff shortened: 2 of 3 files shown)


mystor:
> ```
>     "linker-flavor": "ld.lld",
>     "linker": "rust-lld",
>     "post-link-args": {
>         "ld.lld": [
>             "-z nostart-stop-gc"
>         ]
>     },
> ```

Wouldn't you need to write it more like:

```
"ld.lld": ["-z", "nostart-stop-gc"]
```

because they need to be passed as separate arguments to the linker? Doing that seems to work for me with `rust-lld`.

hawkw:
> > ```
> >     "linker-flavor": "ld.lld",
> >     "linker": "rust-lld",
> >     "post-link-args": {
> >         "ld.lld": [
> >             "-z nostart-stop-gc"
> >         ]
> >     },
> > ```
> 
> Wouldn't you need to write it more like:
> 
> ```
> "ld.lld": ["-z", "nostart-stop-gc"]
> ```
> 
> because they need to be passed as separate arguments to the linker? Doing that seems to work for me with `rust-lld`.

ahahahaha cool i'm a dumbass, thanks. nice catch!

PR #121, opened 2021-10-09
feat: log stuff to the framebuffer, new trace format

hawkw:
this branch logs stuff to the framebuffer.

i'm too lazy to write a good commit message. but like. it works!

image

future work:
- colors!
- draw oops screens on the framebuffer also


hal-core/src/framebuffer/embedded_graphics.rs
```diff
@@ -38,6 +42,7 @@ where
         I: IntoIterator<Item = Pixel<Self::Color>>,
     {
         for Pixel(geometry::Point { x, y }, color) in pixels {
+            // tracing::trace!(x, y, "set_pixel");
```
mystor, line 45: Let's remove the comment :-)

hal-x86_64/src/framebuffer.rs
```diff
@@ -76,6 +76,18 @@ where
...
+
+    fn scroll_vert(&mut self, amount: isize) -> &mut Self {
+        if amount < 0 {
+            todo!("eliza: handle negatives!")
+        }
+        let amount_px = (amount as usize * self.cfg.line_len) * self.cfg.px_bytes;
+        let buf = self.buf.as_mut();
+        let len = buf.len();
+        buf.rotate_left(amount_px);
+        buf[(len - amount_px)..].fill(0);
```
mystor, line 88: This feels like it'd be a bit inefficient to implement, as we're using the more expensive `rotate_left` operation which tries to preserve the rotated out pixels, then immediately clobbering them. Perhaps we should use instead use `copy_within`?

```rust
buf.copy_within(amount_pix.., 0);
buf[(len - amount_pix)..].fill(0);
```

src/arch/x86_64.rs
```diff
@@ -71,19 +71,28 @@ impl BootInfo for RustbootBootInfo {
...
     }
 
     fn subscriber(&self) -> Option<tracing::Dispatch> {
-        Some(tracing::Dispatch::new(
-            hal_x86_64::tracing::Subscriber::default(),
-        ))
+        use mycelium_trace::{writer::MakeWriterExt, Subscriber};
+        if !self.has_framebuffer {
+            // TODO(eliza): we should probably write to just the serial port if
+            // there's no framebuffer...
```
mystor, line 81: That does seem like a good idea. Perhaps follow-up work to make the framebuf optional in the subscriber?

src/arch/x86_64.rs
```diff
@@ -96,6 +105,17 @@ impl BootInfo for RustbootBootInfo {
...
 
 impl RustbootBootInfo {
+    unsafe fn mk_framebuf() -> Framebuffer<'static, spin::MutexGuard<'static, LockedFramebuf>> {
+        let (cfg, buf) = unsafe {
+            // Safety: we can reasonably assume this will only be called
+            // after `arch_entry`, so if we've failed to initialize the
+            // framebuffer...things have gone horribly wrong...
+            FRAMEBUFFER.get_unchecked()
+        };
+        buf.force_unlock();
```
mystor, line 115: It seems quite bad that we can have multiple locks on the same mutex at the same time because we're force-unlocking the mutex here. If we're going to be force-unlocking before we lock, the mutex isn't really doing anything for us.

hal-x86_64/src/framebuffer.rs
hawkw: TIL about `copy_within`, that's much nicer!

src/arch/x86_64.rs
hawkw: whoops, this was supposed to have been removed --- it was added while debugging. my bad!

src/arch/x86_64.rs
hawkw: yeah, i was gonna do that in a subsequent branch

PR #126, opened 2021-10-24
fix(alloc): make add_region loop instead of recursive

hawkw:
this might help avoid stack overflows if we do a lot of splits and end
up with a deeply-nested stack.


mystor (approved):
lgtm

PR #129, opened 2021-11-13
feat(inoculate): `cargo`-style logs with `tracing`

hawkw:
This branch adds a hacked-together `tracing` formatter for
cargo-esque logging in `inoculate`.

Some screenshots:
image
image
image

### DIFF
--- a/inoculate/src/trace.rs
+++ b/inoculate/src/trace.rs
@@ -131,13 +131,13 @@ impl<'styles, 'writer> Visit for Visitor<'styles, 'writer> {
         // If we're writing the first field of the event, either emit cargo
         // formatting, or a level header.
         if self.is_empty {
-            // If the level is `INFO` or `DEBUG` and it has a message that's
+            // If the level is `INFO` and it has a message that's
             // shaped like a cargo log tag, emit the cargo tag followed by the
             // rest of the message.
-            if self.level >= Level::INFO && field.name() == Self::MESSAGE {
+            if self.level == Level::INFO && field.name() == Self::MESSAGE {
                 let message = format!("{:?}", value);
                 if let Some((tag, message)) = message.as_str().split_once(' ') {
-                    if tag.len() <= Self::INDENT && tag.ends_with("ing") || tag.ends_with('d') {
+                    if tag.len() <= Self::INDENT {
                         let tag = tag.to_title_case();
                         let style = match self.level {
                             Level::DEBUG => self.styles.debug,
--- a/inoculate/src/trace.rs
+++ b/inoculate/src/trace.rs
@@ -51,6 +51,11 @@ struct Styles {
     bold: Style,
 }
 
+struct Prefixed<T> {
+    prefix: &'static str,
+    val: T,
+}
+
 impl<S, N> FormatEvent<S, N> for CargoFormatter
 where
     S: Subscriber + for<'a> LookupSpan<'a>,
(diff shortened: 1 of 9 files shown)


inoculate/src/trace.rs
```diff
@@ -0,0 +1,168 @@
...
+{
+    fn format_event(
+        &self,
+        ctx: &FmtContext<'_, S, N>,
+        mut writer: Writer<'_>,
+        event: &Event<'_>,
+    ) -> fmt::Result {
+        let metadata = event.metadata();
+        let level = metadata.level();
+        let color = ColorMode::default().should_color_stdout();
```
mystor, line 27: Why the change to `should_color_stdout()` instead of `should_color_stderr()`? I suppose the old cargo formatter used `stderr` and tracing logs to `stdout`?

inoculate/src/trace.rs
```diff
@@ -0,0 +1,168 @@
...
+    }
+}
+
+impl Visit for Visitor<'_> {
+    fn record_debug(&mut self, field: &Field, value: &dyn fmt::Debug) {
+        if self.is_empty {
+            if self.level >= Level::INFO && field.name() == Self::MESSAGE {
+                let message = format!("{:?}", value);
+                if let Some((tag, message)) = message.as_str().split_once(' ') {
+                    if tag.len() <= Self::INDENT && tag.ends_with("ing") || tag.ends_with('d') {
```
mystor, line 102: this code is definitely not sketchy at all :-)

I suppose doing it with separate fields for the action would be somewhat trickier as `MESSAGE` is usually the first field visited?

inoculate/src/trace.rs
```diff
@@ -0,0 +1,168 @@
...
+                }
+                (Level::TRACE, true) => {
+                    write!(self.writer, "{}{} ", "trace".purple().bold(), ":".bold())
+                }
+                (Level::ERROR, false) => self.writer.write_str("error: "),
+                (Level::WARN, false) => self.writer.write_str("warning: "),
+                (Level::INFO, false) => self.writer.write_str("info: "),
+                (Level::DEBUG, false) => self.writer.write_str("debug: "),
+                (Level::TRACE, false) => self.writer.write_str("trace: "),
+            };
```
mystor, line 146: I wish there was a nicer way to do conditional coloring without writing everything out twice. This is kinda unfortunate :'-(

inoculate/src/trace.rs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        if !self.is_empty {
```
mystor, line 149: Could this be an `else`?

inoculate/src/trace.rs
hawkw: tracing logs to stdout by default; we should probably change it to stderr for consistency with cargo --- i'll do that now

inoculate/src/trace.rs
hawkw: 78cac5fc6ab5b8facd743a10905b2063a16b4ffd should make it a bit nicer --- i forgot that there _was_ a way to do that that i had written previously, lol :/

inoculate/src/trace.rs
hawkw: idk, I kind of like this. alternatively, we could just drop the special casing of "ing" and just always emit the cargo-style formatting for INFO events whose first word is less than the max length. which might actually be nicer, idk.

mystor: I kinda like the idea of requiring a pattern like `Action: message goes here`, where a `:` after the first word is stripped and used to determine that it was intended to be a cargo style info event, but I don't feel a strong opinion about it.

PR #132, opened 2022-01-01
chore: update to nightly-2021-12-31

hawkw:
This branch updates mycelium to build with the latest nightly, Rust
2021-12-31. This was a pretty easy upgrade. I made the following
changes:

- import the `asm!` macro from `core::arch::asm`, which is where it
  lives now
- stick a bunch of `must_use` attributes on basically every function
  that returns a `Self` and doesn't mutate the receiver (clippy whines
  about this now)
- update dependencies


hawkw:
oh, nvm, this doesn't actually work because `wasmi` is invoking `asm!` from the wrong place, apparently, so that'll have to be fixed upstream.

PR #161, opened 2022-05-23
Add cache padding inhibitor

jamesmunns:
As requested!

If this gets more complicated, it might be worth bringing in the `cfg_if` crate.

### DIFF
--- a/cordyceps/Cargo.toml
+++ b/cordyceps/Cargo.toml
@@ -18,6 +18,10 @@ categories = ["data-structures", "no-std"]
 default = []
 alloc = []
 std = ["alloc"]
+
+# Inhibits cache padding for the `CachePadded` struct used for many linked list pointers
+# When this feature is NOT enabled, the size will be determined base on target platform.
+# See the documentation of `CachePadded` for more details.
 no-cache-pad = []
 
 [dev-dependencies]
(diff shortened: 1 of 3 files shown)


cordyceps/Cargo.toml
```diff
@@ -18,6 +18,7 @@ categories = ["data-structures", "no-std"]
 default = []
 alloc = []
 std = ["alloc"]
+no-cache-pad = []
```
hawkw, line 21: maybe worth a comment explaining what this does here? and/or documenting it in the `lib.rs` docs? but, we don't currently have a list of other feature flags, so it would be fine to add that in a separate PR...

hawkw (approved):
looks good to me --- one note on docs, but not a blocker

cordyceps/README.md
```diff
@@ -44,4 +44,15 @@ ascomycete fungi that's (in)famous for its [intrusive behavior][cordyceps].
 
 [Mycelium]: https://github.com/hawkw/mycelium
 [intrusive]: https://www.boost.org/doc/libs/1_45_0/doc/html/intrusive/intrusive_vs_nontrusive.html
-[cordyceps]: https://en.wikipedia.org/wiki/Cordyceps#Biology
\ No newline at end of file
+[cordyceps]: https://en.wikipedia.org/wiki/Cordyceps#Biology
+
+## features
+
+The following features are available (this list is incomplete; you can help by [expanding it].)
```
hawkw, line 51: i love this

PR #163, opened 2022-05-25
Const constructor method

jamesmunns:


cordyceps/rust-toolchain.toml
```diff
@@ -0,0 +1,10 @@
+[toolchain]
```
jamesmunns, line 1: So, this is allowed on the current stable and nightly, but not on the project level nightly. Added a file here in case you don't want to rev the whole project.

This is due to the const bounds on the const constructor method. I can also feature flag that out for easier compat.

hawkw: happy to bump nightly if nothing’s broken (and if anything is broken, still happy to bump nightly but i’d have to fix that first)

cordyceps/src/mpsc_queue.rs
```diff
@@ -394,10 +396,22 @@ impl<T: Linked<Links<T>>> MpscQueue<T> {
             head: CachePadded(AtomicPtr::new(ptr)),
             tail: CachePadded(UnsafeCell::new(ptr)),
             has_consumer: CachePadded(AtomicBool::new(false)),
+            stub_is_static: false,
             stub,
         }
     }
 
+    pub const fn new_with_static_stub(stub: &'static T) -> Self {
```
hawkw, line 404: could we maybe just stick this in an impl block without bounds on it? it would be fine to remove the type-level bounds.

cordyceps/rust-toolchain.toml
jamesmunns: Pushed a commit updating the main toolchain version. I'm not sure if there is still the existing cordyceps failures we talked about present.

cordyceps/src/mpsc_queue.rs
hawkw: wait never mind it’s  needed for the drop impl so we can’t get rid of it. welp. if i publish this to crates.io we could maybe `autocfg` this constructor out so it can build on stable?

jamesmunns: This builds on the current stable (1.61 stabilized trait bounds on consts)

jamesmunns:
Also worth noting, if you are in debug mode, AND you use a static constructor, AND you forget to use the `new_stub` method for creating the `Links` structure, you will encounter a failed debug assert.

If any of those are not true, the end user will not see any difference.

I can't find a way to avoid this, as the trait methods (and any unsafe tricks I could think of to do an assert) wouldn't work.

hawkw:
> Also worth noting, if you are in debug mode, AND you use a static constructor, AND you forget to use the `new_stub` method for creating the `Links` structure, you will encounter a failed debug assert.
> 
> 
> 
> If any of those are not true, the end user will not see any difference.
> 
> 
> 
> I can't find a way to avoid this, as the trait methods (and any unsafe tricks I could think of to do an assert) wouldn't work.

we could probably just get rid of the `is_stub` field, or make it `#[cfg(test)]`-only rather then `#[cfg(debug_assertions)]`...

hawkw:
my preference, fwiw, would be to do the nightly version bump for the whole project, provided the kernel still builds successfully on the latest nightly. 

if it _doesn't_, though, it's certainly not your responsibility to fix that, and i'd be fine with adding a separate `rust-toolchain.toml` for just Cordyceps temporarily (the current approach), until i can figure out the update...

jamesmunns:
> my preference, fwiw, would be to do the nightly version bump for the whole project, provided the kernel still builds successfully on the latest nightly.

I did this in commit [4dafbe6](https://github.com/hawkw/mycelium/pull/163/commits/4dafbe6e94820f0f5a77d90cc68470eada286eb1). AFAICT everything still passes.

> we could probably just get rid of the is_stub field, or make it #[cfg(test)]-only rather then #[cfg(debug_assertions)]...

I don't have a preference, I'll let you change that in a later PR I think, right now I've "worked around" it by documenting it and providing examples :D

jamesmunns:
So, I reran the miri tests locally, and they DO pass, but it's worth noting that shortly before exiting, it was using **11GiB** of resident memory.

I wonder if the CI test is being killed by OOM, causing a panic and test failure, as there is no indicator of what condition failed. It might be necessary to rework the test cases to see where it is going, and if the peak memory can be reduced.

Test run from my local machine:

```
➜  mycelium git:(static-stub) ./bin/miri -p cordyceps
++ dirname ./bin/miri
+ cd ./bin/..
+ MIRIFLAGS='-Zmiri-disable-isolation -Zmiri-tag-raw-pointers '
+ PROPTEST_CASES=10
+ cargo miri test --lib -p cordyceps
    Updating crates.io index
...
   Compiling cordyceps v0.1.0 (/home/james/personal/mycelium/cordyceps)
    Finished test [optimized + debuginfo] target(s) in 13.69s
     Running unittests src/lib.rs (target/miri/x86_64-unknown-linux-gnu/debug/deps/cordyceps-5a66377296e975c6)

running 15 tests
test list::tests::const_new ... ok
test list::tests::fuzz_linked_list ... ok
test list::tests::push_and_drain ... ok
test list::tests::push_pop_push_pop ... ok
test list::tests::remove_by_address::first ... ok
test list::tests::remove_by_address::last ... ok
test list::tests::remove_by_address::last_middle ... ok
test list::tests::remove_by_address::middle ... ok
test list::tests::remove_by_address::missing ... ok
test mpsc_queue::tests::basically_works ... warning: thread support is experimental and incomplete: weak memory effects are not emulated.

ok
test mpsc_queue::tests::basically_works_const ... ok
test mpsc_queue::tests::dequeue_empty ... ok
test mpsc_queue::tests::enqueue_dequeue ... ok
test mpsc_queue::tests::try_dequeue_busy ... ok
test mpsc_queue::tests::try_dequeue_empty ... ok

test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

hawkw (approved):
overall, this looks good to me! i commented on some very minor nits, mostly docs-related.

cordyceps/src/mpsc_queue.rs
```diff
@@ -415,10 +417,96 @@ impl<T: Linked<Links<T>>> MpscQueue<T> {
...
+    ///
+    /// ## Usage notes
+    ///
+    /// Unlike `new()` or `new_with_stub()`, the `stub` item will NOT be
+    /// dropped when the `MpscQueue` is dropped. This is fine if you are
+    /// ALSO statically creating the `stub`, however if it is necessary to
+    /// recover that memory after the MpscQueue has been dropped, that will
+    /// need to be done by the user manually.
+    ///
+    /// ## SAFETY
```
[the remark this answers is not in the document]
hawkw: (there's a Clippy lint for `pub unsafe fn`s without a safety section in their docs, and it specifically looks for `# Safety`, so this is failing that lint)

cordyceps/src/mpsc_queue.rs
```diff
@@ -415,10 +417,96 @@ impl<T: Linked<Links<T>>> MpscQueue<T> {
...
+    /// # }
+    ///
+    /// static STUB_ENTRY: Entry = Entry {
+    ///     links: mpsc_queue::Links::<Entry>::new_stub(),
+    ///     val: 0
+    /// };
+    ///
+    ///
+    /// // SAFETY: The STUB_ENTRY is only ever used for one MPMC queue
+    /// static MPMC: MpscQueue<Entry> = unsafe { MpscQueue::new_with_static_stub(&STUB_ENTRY) };
```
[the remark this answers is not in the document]
hawkw: update: i tested this out locally and it does work, so I'm going to make that change as well.

cordyceps/src/mpsc_queue.rs
```diff
@@ -341,6 +341,8 @@ pub struct MpscQueue<T: Linked<Links<T>>> {
     /// new consumer.
     has_consumer: CachePadded<AtomicBool>,
 
+    stub_is_static: bool,
```
hawkw, line 344:
```suggestion
    /// If the stub node is in a `static`, we cannot drop it when the 
    /// queue is dropped.
    stub_is_static: bool,
```

cordyceps/src/mpsc_queue.rs
```diff
@@ -1283,6 +1395,99 @@ mod tests {
...
+                Err(e) => {
+                    println!("recv error {:?}", e);
+                    thread::yield_now();
+                }
+            }
+        }
+
+        for thread in threads {
+            thread.join().unwrap();
+        }
```
hawkw, line 1441: it occurs to me that the main body of most of these tests could probably be factored out, but we can do that later...

PR #164, opened 2022-05-25
Isolate mpmc test fix

jamesmunns:


hawkw (changes requested):
thanks, good catch!

i think we may need to make the same change here https://github.com/hawkw/mycelium/blob/3126dabe4c3ddc52319007e153bfa325cd594be2/cordyceps/src/mpsc_queue.rs#L270-L274
and here
https://github.com/hawkw/mycelium/blob/3126dabe4c3ddc52319007e153bfa325cd594be2/cordyceps/src/mpsc_queue.rs#L123-L131
which i think can also suffer from the same issue?

hawkw (approved):
i went ahead and fixed the other tests. will merge pending CI!

PR #175, opened 2022-05-30
wip: abstraction for custom box types for tasks

hawkw:

### DIFF
--- a/async/src/task.rs
+++ b/async/src/task.rs
@@ -16,12 +16,24 @@ use core::{
     task::{RawWaker, RawWakerVTable},
 };
 
+mod allocation;
 mod state;
+pub use self::allocation::Allocation;
 
 use self::state::StateVar;
 
 #[derive(Debug)]
-pub(crate) struct TaskRef(NonNull<Header>);
+pub struct TaskRef(NonNull<Header>);
+
+#[repr(C)]
+pub struct Task<S, F: Future> {
+    header: Header,
+
+    scheduler: S,
+    inner: UnsafeCell<Cell<F>>,
+}
+
+pub trait Allocation {}
 
 #[repr(C)]
 #[derive(Debug)]
@@ -30,14 +42,6 @@ pub(crate) struct Header {
     state: StateVar,
     // task_list: list::Links<TaskRef>,
     vtable: &'static Vtable,
-}
-
-#[repr(C)]
-struct Task<S, F: Future> {
-    header: Header,
-
-    scheduler: S,
-    inner: UnsafeCell<Cell<F>>,
 }
 
 enum Cell<F: Future> {
--- a/async/Cargo.toml
+++ b/async/Cargo.toml
@@ -9,6 +9,10 @@ license = "MIT"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+default = ["alloc"]
+alloc = [""]
+
 [dependencies]
 mycelium-util = { path = "../util" }
 mycotest = { path = "../mycotest", default-features = false }
@@ -25,3 +29,7 @@ loom = { version = "0.5.5", features = ["futures"] }
 tracing_01 = { package = "tracing", version = "0.1", default_features = false }
 tracing_subscriber_03 = { package = "tracing-subscriber", version = "0.3.11", features = ["fmt"] }
 futures-util = "0.3"
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
--- a/async/src/lib.rs
+++ b/async/src/lib.rs
@@ -1,4 +1,7 @@
 #![cfg_attr(not(test), no_std)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
+#![cfg_attr(docsrs, doc(cfg_hide(docsrs)))]
+#[cfg(feature = "alloc")]
 extern crate alloc;
 
 #[macro_use]
--- a/async/src/task/allocation.rs
+++ b/async/src/task/allocation.rs
@@ -1,0 +1,23 @@
+use super::Task;
+use core::ptr::NonNull;
+
+pub trait HeapTask<S, F> {
+    pub fn into_raw(self) -> NonNull<S, F>;
+    pub fn from_raw(ptr: NonNull<S, F>) -> Self;
+}
+
+feature! {
+    #![feature = "alloc"]
+    use alloc::boxed::Box;
+
+    impl<S, F> HeapTask<S, F> for Box<Task<S, F>> {
+        pub fn into_raw(self) -> NonNull<S, F> {
+            crate::util::non_null(Box::into_raw(self))
+        }
+
+        pub fn from_raw(ptr: NonNull<S, F>) -> Self {
+            unsafe { Box::from_raw(ptr.as_ptr()) }
+        }
+    }
+
+}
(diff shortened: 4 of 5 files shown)


jamesmunns:
@hawkw started working on this, got it to a point where it *compiles*, but doesn't really do much, because the `spawn` trait essentially requires allocation.

I also don't *love* the third generic on the `Task` item (and dummy types), but I'm not sure how else to do it without incurring type recursion. I figure we can use this until we think of something better.

Let me know if you don't want me touching your `eliza/` branches, and we can force revert and I'll make a PR to this PR.

async/src/scheduler.rs
```diff
@@ -1,99 +1,135 @@
...
-impl<F: Future> Spawn<F> for Scheduler {
-    /// Spawns `future` as a new task on this executor.
-    fn spawn(&self, future: F) {
-        Scheduler::spawn(self, future)
-    }
-}
+pub use static_scheduler::*;
 
-// === impl StaticScheduler ===
+mod static_scheduler {
```
jamesmunns, line 64: By the way, I added these mostly to keep things organize for myself, if you want me to revert this I can.

jamesmunns:
I think this is ready for a coarse "design review", to see if you like this conceptually, or if you have any more elegant ways of doing things (particularly with the `STO: Storage<S, F>` trait bounds).

If it looks conceptually good, I can start writing docs and tests that cover the new additions.

hawkw:
> Let me know if you don't want me touching your `eliza/` branches, and we can force revert and I'll make a PR to this PR.

It's totally fine as long as I'm not actively working on this branch (which, in this case, I'm not). If I was still pushing changes to this, it would be better to make your branch off of it, but in this case, it's totally fine. :)

hawkw (commented):
> I think this is ready for a coarse "design review", to see if you like this conceptually, or if you have any more elegant ways of doing things (particularly with the `STO: Storage<S, F>` trait bounds).
> 
> If it looks conceptually good, I can start writing docs and tests that cover the new additions.

Overall, this approach seems fine to me; I agree that the additional generic everywhere is kind of unfortunate, but I can't think of a better way to do it either. I have a slight preference for not making the `task::Header` type public _just_ for `new_stub`, but if that's the only way to do it, that's fine too.

I commented on some minor style nitpicks/suggestions inline. I think this design looks good overall, though, so let's go ahead and start getting it ready to merge!

maitake/src/lib.rs
```diff
@@ -1,6 +1,9 @@
 #![cfg_attr(docsrs, doc = include_str!("../README.md"))]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
 #![cfg_attr(not(test), no_std)]
+#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
+#![cfg_attr(docsrs, doc(cfg_hide(docsrs)))]
```
hawkw, line 5: looks like there was some duplication in the merge here:
```suggestion
#![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
#![cfg_attr(docsrs, doc(cfg_hide(docsrs)))]
#![cfg_attr(not(test), no_std)]
```

maitake/Cargo.toml
```diff
@@ -40,4 +44,4 @@ tracing_subscriber_03 = { package = "tracing-subscriber", version = "0.3.11", fe
 
 [package.metadata.docs.rs]
 all-features = true
-rustdoc-args = ["--cfg", "docsrs"]
```
jamesmunns, line 43: @hawkw you need to configure your editor to add trailing newlines on save :p

maitake/src/scheduler.rs
```diff
@@ -102,19 +115,18 @@ impl Core {
...
         }
     }
 
-    fn spawn_static(&'static self, future: impl Future) {
-        self.schedule(TaskRef::new(self, future));
-    }
-
-    fn spawn_arc(this: &Arc<Self>, future: impl Future) {
-        this.schedule(TaskRef::new(this.clone(), future));
+    const unsafe fn new_with_static_stub(stub: &'static Header) -> Self {
```
hawkw, line 126: needs to be
```suggestion
    #[cfg(not(loom))]
    const unsafe fn new_with_static_stub(stub: &'static Header) -> Self {
```
since `loom`'s `AtomicPtr::new` can't be `const`: https://github.com/hawkw/mycelium/runs/6677548757?check_suite_focus=true

maitake/src/util.rs
```diff
@@ -52,6 +52,20 @@ macro_rules! fmt_bits {
     };
 }
 
+#[macro_export]
```
hawkw, line 55: added this accidentally:
```suggestion
```

hawkw (commented):
CI is upset about the static stub stuff in the loom build: https://github.com/hawkw/mycelium/runs/6677548757?check_suite_focus=true

async/src/task.rs
```diff
@@ -1,53 +1,76 @@
...
+pub struct Header {
     run_queue: mpsc_queue::Links<Header>,
     state: StateCell,
     // task_list: list::Links<TaskRef>,
     vtable: &'static Vtable,
 }
 
 #[repr(C)]
-struct Task<S, F: Future> {
+pub struct Task<S, F: Future, STO> {
```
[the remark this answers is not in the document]
jamesmunns: Yeah, I don't think this is possible.

jamesmunns:
So, regarding `Header`:

> I have a slight preference for not making the task::Header type public just for new_stub, but if that's the only way to do it, that's fine too.

So, to address this, it would be a bit of a circular problem.

To create a new `StaticScheduler`, you need to give it a "stub task" to start with. But the `Task` type includes a reference (or Arc) of the Scheduler.

My *current* approach cheats by not actually making a task, only a header, since it knows it will only ever be used for the Mpmc queue.

I'll go with the only other approach I can think of: making an opaque "TaskStub" type.

maitake/src/task/storage.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+feature! {
+    #![feature = "alloc"]
+    use alloc::boxed::Box;
+
+    /// A type representing [`Box`] storage of a task
+    ///
+    /// [`Box`]: alloc::boxed::Box
+    pub struct BoxStorage;
```
jamesmunns, line 58: @hawkw should I add an internal private type to make this externally unconstructable?

I don't think you can *do* anything with the type anyway.

jamesmunns:
Not sure why that netlify task is stalled?

maitake/src/task/storage.rs
```diff
@@ -0,0 +1,74 @@
...
+///
+/// This trait is exposed publicly to allow for end users to implement
+/// it in their own heap allocation types, if they are not specifically
+/// a [`Box`].
+///
+/// This trait is ONLY appropriate for heap allocation types that represent
+/// exclusive ownership of the contained data, as it may be mutated while
+/// in pointer form. For example, the `Box` type would match this guarantee,
+/// but the `Arc` type would not, as it allows for shared access, and NOT
+/// exclusive mutable access.
```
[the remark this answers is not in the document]
hawkw: (also, i'm not actually sure if this is necessarily a required invariant, since the task type is interior-ly mutable; the future lives in an `UnsafeCell` and only polling or dropping the task will mutate the `UnsafeCell`; the atomic state variable ensures mutual exclusion of the mutable part of the task. an earlier design used `Arc`s rather than `Box` but i abandoned that approach because i wanted to roll the ref count into the same atomic as the task's state flags...)

hawkw (commented):
some last final suggestions, and then i think this should be good to go! excited for this change!

hawkw:
i made a handful of tiny changes, and i'll be happy to merge this once CI passes!

PR #177, opened 2022-05-30
feat(async): implement task state transitions

hawkw:
this branch adds a proper implementation of task state transitions to
the async crate's task system. right now, the set of state transitions
is fairly small, but this will become more complex if we add
`JoinHandle`s or a similar abstraction for awaiting a task's completion.
also, i added tests exercising `wake_by_val` wakeups, and, it turns out,
we were leaking tasks in this case. so that's fixed now.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -532,6 +532,7 @@ dependencies = [
  "cordyceps",
  "futures-util",
  "loom",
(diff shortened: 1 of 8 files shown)


hawkw:
cc @jamesmunns you may be interested in this diff...

PR #180, opened 2022-05-31
maitake: change name to `maitake`, add docs

hawkw:


maitake/README.md
```diff
@@ -0,0 +1,63 @@
...
+[sponsor-url]: https://github.com/sponsors/hawkw
+[maitake-wiki]: https://en.wikipedia.org/wiki/Grifola_frondosa
+
+## what is it?
+
+This library is a collection of modular components for building a Rust
+async runtime based on [`core::task`] and [`core::future`], with a focus on
+supporting `#![no_std]` projects.
+
+Unlike other async runtime implementations, `maitake` does *not* provide a
```
jamesmunns, line 35: Compare to `smol` perhaps?

maitake/README.md
```diff
@@ -0,0 +1,63 @@
...
+`maitake` was initially designed for use in the [mycelium] and [mnemOS]
+operating systems, but may be useful for other projects as well.
+
+[`core::task`]: https://doc.rust-lang.org/stable/core/task/index.html
+[`core::future`]: https://doc.rust-lang.org/stable/core/future/index.html
+[task system]: task
+[scheduling]: scheduler
+[wait]: wait
+[mycelium]: https://github.com/hawkw/mycelium
+[mnemOS]: https://github.com/jamesmunns/pellegrino/tree/main/firmware/kernel
```
jamesmunns, line 51:
```suggestion
[mnemOS]: https://mnemos.jamesmunns.com
```

maitake/README.md
hawkw: oh good point, smol is in _some_ ways similar, but the primitives it exposes are somewhat higher-level

jamesmunns: Gotcha, you know better than me! I just knew that `smol` had been re-used in multiple executors as sort of a "core" building block.

hawkw: i might add a whole "comparison to other crates" section in a follow-up PR?

jamesmunns (approved):
Overall looks good to me! Made two small suggestions, feel free to merge when you're happy.

:partying_face:

maitake/README.md
jamesmunns: Sounds good! The natural comparisons would be embassy (also no_std focused, but different), smol (as another building-block-ish crate), and tokio/async-std (as full-featured, but std-focused runtimes/ecosystems)

hawkw: > Gotcha, you know better than me! I just knew that `smol` had been re-used in multiple executors as sort of a "core" building block.

yeah, it's still a slightly higher level primitive, i think --- this crate is more comparable to something like [`async-task`](https://docs.rs/async-task/latest/async_task/) which is a dependency of `smol`.

jamesmunns: Specifically for embassy, maybe focus on the fact that maitake is optionally-alloc, and supports dynamic task spawning

PR #183, opened 2022-06-01
Fix links in Maitake's readme

jamesmunns:


jamesmunns:
Ah, gotcha! I was just browsing through on GH (wondering about the task primitives) and saw the links were broken.

I don't have any good solutions, modulo not using `include_str` for the top level README :/

jamesmunns:
Closing this for now. We might consider splitting the top level docs and readme content before the initial release.

PR #191, opened 2022-06-03
feat(maitake): add `WaitQueue`

hawkw:
this branch adds a `WaitQueue` type in `maitake::wait`. this allows
multiple tasks to wait for an event, and be woken either one at a time
(in FIFO fashion), or all at once.

there are additional APIs we could add here, but this is a good starting
place. there probably should be docs examples but i'm lazy.

### DIFF
--- a/maitake/src/wait/queue.rs
+++ b/maitake/src/wait/queue.rs
@@ -326,6 +326,7 @@ impl WaitQueue {
         state = self.load();
 
         if let Some(waker) = self.wake_locked(&mut *queue, state) {
+            drop(queue);
             waker.wake();
         }
     }
@@ -463,7 +464,7 @@ impl WaitQueue {
         if test_dbg!(state) != State::Waiting {
             // if there are no longer any queued tasks, try to store the
             // wakeup in the queue and bail.
-            if let Err(actual) = self.compare_exchange(curr, curr.with_state(State::Waiting)) {
+            if let Err(actual) = self.compare_exchange(curr, curr.with_state(State::Woken)) {
                 debug_assert!(actual.get(QueueState::STATE) != State::Waiting);
                 self.store(actual.with_state(State::Woken));
             }
--- a/maitake/src/wait/queue.rs
+++ b/maitake/src/wait/queue.rs
@@ -375,8 +375,13 @@ impl WaitQueue {
     /// synchronization primitives or resources: when an event makes a resource
     /// permanently unavailable, the queue can be closed.
     pub fn close(&self) {
+        let state = self.state.fetch_or(State::Closed as u8 as usize, SeqCst);
+        let state = test_dbg!(QueueState::from_bits(state));
+        if state.get(QueueState::STATE) != State::Waiting {
+            return;
+        }
+
         let mut queue = self.queue.lock();
-        test_dbg!(self.state.fetch_or(State::Closed as u8 as usize, SeqCst));
 
         // TODO(eliza): wake outside the lock using an array, a la
         // https://github.com/tokio-rs/tokio/blob/4941fbf7c43566a8f491c64af5a4cd627c99e5a6/tokio/src/sync/batch_semaphore.rs#L277-L303
@@ -479,12 +484,6 @@ impl WaitQueue {
         }
 
         Some(waker)
-    }
-}
-
-impl Drop for WaitQueue {
-    fn drop(&mut self) {
-        self.close();
     }
 }
 
@@ -679,6 +678,7 @@ impl Waiter {
         // node in the queue.
         if Waiter::with_node(ptr, &mut waiters, |node| matches!(&node.waker, Wakeup::One)) {
             if let Some(waker) = queue.wake_locked(&mut waiters, state) {
+                drop(waiters);
                 waker.wake()
             }
         }
--- a/maitake/src/wait/queue.rs
+++ b/maitake/src/wait/queue.rs
@@ -45,16 +45,105 @@ mod tests;
 /// primitive on its own: sometimes, you just need to have a bunch of tasks wait
 /// for something and then wake them all up.
 ///
+/// # Examples
+///
+/// Waking a single task at a time by calling [`wake`][wake]:
+///
+/// ```
+/// use std::sync::Arc;
+/// use maitake::{scheduler::Scheduler, wait::WaitQueue};
+///
+/// const TASKS: usize = 10;
+///
+/// // In order to spawn tasks, we need a `Scheduler` instance.
+/// let scheduler = Scheduler::new();
+///
+/// // Construct a new `WaitQueue`.
+/// let q = Arc::new(WaitQueue::new());
+///
+/// // Spawn some tasks that will wait on the queue.
+/// for _ in 0..TASKS {
+///     let q = q.clone();
+///     scheduler.spawn(async move {
+///         // Wait to be woken by the queue.
+///         q.wait().await.expect("queue is not closed");
+///     });
+/// }
+///
+/// // Tick the scheduler once.
+/// let tick = scheduler.tick();
+///
+/// // No tasks should complete on this tick, as they are all waiting
+/// // to be woken by the queue.
+/// assert_eq!(tick.completed, 0, "no tasks have been woken");
+///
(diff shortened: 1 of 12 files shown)


maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+        let state = self.load();
+
+        match state.get(QueueState::STATE) {
+            // if the queue is closed, bail.
+            State::Closed => return,
+
+            // if there are no waiters in the queue, increment the number of
+            // `wake_all` calls and return.
+            State::Woken | State::Empty => {
+                self.state.fetch_add(QueueState::ONE_WAKE_ALL, SeqCst);
```
jamesmunns, line 346: Could this be made AcqRel?

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+                return;
+            }
+            State::Waiting => {}
+        }
+
+        // okay, we actually have to wake some stuff.
+
+        // TODO(eliza): wake outside the lock using an array, a la
```
jamesmunns, line 354: Can this todo be closed? I don't really track...

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// Close the queue, indicating that it may no longer be used.
+    ///
+    /// Once a queue is closed, all [`wait`] calls (current or future) will
+    /// return an error.
+    ///
+    /// This method is generally used when implementing higher-level
+    /// synchronization primitives or resources: when an event makes a resource
+    /// permanently unavailable, the queue can be closed.
+    pub fn close(&self) {
+        let state = self.state.fetch_or(State::Closed as u8 as usize, SeqCst);
```
jamesmunns, line 379: Might wanna put a big ol' warning flag on the State variable that this is dependent on "Closed" being "all bits set" of the relevant state fields

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+}
+
+// === impl Waiter ===
+
+impl Waiter {
+    /// Wake the task that owns this `Waiter`.
+    ///
+    /// # Safety
+    ///
+    /// This is only safe to call while the list is locked. The dummy `_list`
```
jamesmunns, line 498: The comment say it's a dummy, but `list` is used inside the function

maitake/src/wait/queue/tests.rs
```diff
@@ -0,0 +1,298 @@
...
+
+#[cfg(all(not(loom), feature = "alloc"))]
+mod alloc {
+    use super::*;
+    use crate::loom::sync::Arc;
+    use crate::scheduler::Scheduler;
+    use core::sync::atomic::{AtomicUsize, Ordering};
+
+    #[test]
+    fn wake_all() {
```
jamesmunns, line 11: This would be awesome to steal as an example at the top level docs.

jamesmunns (commented):
Added some high level comments - I admit to not totally grokking how everything works yet, but I'm going to go try and integrate this branch into mnemos now, so I might have more meaningful feedback after that.

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// # Safety
+    ///
+    /// This *must* be the first field in the struct in order for the `Linked`
+    /// impl to be sound.
+    links: list::Links<Waiter>,
+
+    /// The node's waker
+    waker: Wakeup,
+
+    // /// Optional user data.
```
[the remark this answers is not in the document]
hawkw: this is commented out because i punted on implementing user data in this branch, i should probably just delete it.

maitake/src/wait/queue.rs
hawkw: this is a thing i still want to do in a follow-up PR.

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+    /// When in this state, the node is **not** part of the linked list, and
+    /// can be dropped without removing it from the list.
+    Woken,
+}
+
+/// A waiter node which may be linked into a wait queue.
+#[derive(Debug)]
+#[repr(C)]
+#[pin_project]
+struct Waiter {
```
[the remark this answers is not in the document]
hawkw: i don't really think this is a `Header`; in the case of the task header, the naming is because the header is a small amount of data stored before the task's "body" (the future/output). in this case, the `Waiter` type is the entirety of the data in the linked list...

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,790 @@
...
+    queue: &'a WaitQueue,
+
+    /// Entry in the wait queue linked list.
+    #[pin]
+    waiter: Waiter,
+}
+
+/// The state of a [`Waiter`] node in a [`WaitQueue`].
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+enum WaitState {
```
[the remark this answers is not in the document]
hawkw: 30/60 bits is definitely enough, because it always comes from the value stored in `QueueState`, which is also missing 2 bits...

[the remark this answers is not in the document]
hawkw: done in 8cc540eab56064e313b9c5b533869f436c764ef1

hawkw:
@jamesmunns i think i've addressed most of your review feedback; if you're still working on integrating this in mnemOS, i'd be happy to wait for that, but i  have a slight preference for merging this now and addressing any additional issues that come up in a follow-up, if that's okay?

jamesmunns:
Misc note (can totally be a future PR): It might be nice to be able to modify the lock behavior somehow.

For me, any lock contention is likely to be a programming bug, so `.try_lock().unwrap()` is likely to make a "louder"/easier to catch case for me.

jamesmunns:
> @jamesmunns i think i've addressed most of your review feedback; if you're still working on integrating this in mnemOS, i'd be happy to wait for that, but i have a slight preference for merging this now and addressing any additional issues that come up in a follow-up, if that's okay?

It's integrated and working, no complaints from use! I'll take a quick scan of the recent changes and r+ it.

hawkw:
> 



> Misc note (can totally be a future PR): It might be nice to be able to modify the lock behavior somehow.
> 
> For me, any lock contention is likely to be a programming bug, so `.try_lock().unwrap()` is likely to make a "louder"/easier to catch case for me.

hmm, this isn't a bad idea, but it might be kind of painful to figure out a good interface for overriding lock behavior...could be worth opening a separate issue for?

maitake/src/wait/queue.rs
```diff
@@ -0,0 +1,927 @@
...
+    pub fn new() -> Self {
+        Self {
+            state: CachePadded::new(AtomicUsize::new(State::Empty.into_usize())),
+            queue: Mutex::new(List::new()),
+        }
+    }
+
+    /// Wake the next task in the queue.
+    ///
+    /// If the queue is empty, a wakeup is stored in the `WaitQueue`, and the
```
jamesmunns, line 362: Two things I don't grok:

What happens if I call "wake" a bunch with an empty queue?

How come this stores a "wakeup" in the queue, rather than keeping it at the "empty" state?

jamesmunns (approved):
Overall LGTM! Added one "I don't understand" question, but that's mostly on me, unless you want to update the docs to explain.

maitake/src/wait/queue.rs
hawkw: tried changing all atomic ops to AcqRel/Acquire, and one of the loom tests broke; i think there are at least some cases where SeqCst is necessary (perhaps to synchronize with the mutex?). will investigate further in a follow up.

maitake/src/wait/queue.rs
hawkw: it's necessary to store a single wakeup in the queue because a task may be attempting to enqueue itself *while* `wake` is called, which would result in a "lost" wake. multiple calls to `wake` on an empty queue will not store additional wakeups; storing a count of wakeups in the queue seems like a job for a semaphore, which i'd also like to add to `maitake`.

Issue #186, opened 2022-06-03
cordyceps: pop_front/push_back for List


hawkw:
right now the `cordyceps::List` type only has queue-like `push_front` and `pop_back` methods. as a doubly-linked list, it stands to reason that we could also add `pop_front` and `push_back`, allowing it to be used as a double-ended queue or stack.

hawkw:
(i currently have no particular need for this, but it Would Be Nice To Have)

PR #198, opened 2022-06-05
feat(cordyceps): added new push_back and pop_front methods to list

Tyrannican:
This PR adds some nice-to-have methods as detailed in [this issue](https://github.com/hawkw/mycelium/issues/186).

Changes:
* New `push_back` method
* New `pop_front` method
* Tests for each method

Small change but hope it's good :D

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -665,6 +665,9 @@ mod tests {
         let f = list.pop_back().unwrap();
         assert_eq!(5, f.val);
 
+        assert!(list.is_empty());
+        assert!(list.pop_back().is_none());
+
         list.assert_valid();
     }
 
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -158,6 +158,49 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         // tracing::trace!(?self, "push_front: pushed");
     }
 
+    /// Appends an item to the tail of the list
+    pub fn push_back(&mut self, item: T::Handle) {
+        let ptr = T::into_ptr(item);
+        // tracing::trace!(?self, ?ptr, "push_back");
+        assert_ne!(self.tail, Some(ptr));
+        unsafe {
+            T::links(ptr).as_mut().set_next(None);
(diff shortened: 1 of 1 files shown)


cordyceps/src/list.rs
```diff
@@ -158,6 +158,49 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         // tracing::trace!(?self, "push_front: pushed");
     }
 
+    /// Appends an item to the tail of the list
+    pub fn push_back(&mut self, item: T::Handle) {
+        let ptr = T::into_ptr(item);
+        // tracing::trace!(?self, ?ptr, "push_back");
```
hawkw, line 164: my guess is that these are here because they're in the `push_back` and `pop_front` methods, but because they're commented out, i have a slight preference for not adding them in new code.

we should maybe eventually add a feature flag to control whether tracing is enabled in `cordyceps`... but we should probably do that in a separate branch.

hawkw (commented):
thanks for the PR! i have a couple of very minor suggestions:

- the RustDoc for the `List` type references using the list as a FIFO using `push_front` and `pop_back`, it would be nice to add something saying that it can also be used as a stack or double-ended queue using these methods
- thanks for adding tests! it could be good to add these methods to the `proptest` `fuzz_linked_list` test as well, but if that's too much work, it would be fine to do that in a follow-up branch. 

thanks again for the PR!

cordyceps/src/list.rs
Tyrannican: Removed for this PR! :D

Tyrannican:
* Added the tests to the `proptest` `fuzz_linked_list`
* Added documentation to detail the List data structure can be used as a stack or Doubly Linked List

hawkw (approved):
this looks great, i'm going to merge it once CI has run! thanks again!

hawkw (approved):
this looks great to me, i'm going to merge as soon as CI is done. thanks again!

hawkw:
oh, looks like `rustfmt` also needs to be run: https://github.com/hawkw/mycelium/runs/6747421224?check_suite_focus=true

mind fixing that? thanks!

Tyrannican:
Done!

PR #209, opened 2022-06-08
fix(cordyceps): pin `list::IterMut` items

hawkw:
Currently, there is a potential safety issue when using
`cordyceps::list::IterMut`: the iterator's item type is `&mut T`.
`&mut T`s may be `mem::replace`d or `mem::take`en, moving them out of
the list to a different memory location. This would invalidate any
pointers in the list that point to that node.

Therefore, to ensure that the mutable iterator cannot corrupt the list
(potentially leading to dangling pointers or use-after-frees), this PR
changes `list::IterMut`'s `Item` type from `&'list mut T` to 
`Pin<&'list mut T`. This ensures that the iterator cannot be used to 
corrupt the list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,7 @@ name = "cordyceps"
 version = "0.2.0"
 dependencies = [
  "loom",
+ "pin-project",
  "proptest",
  "tracing 0.1.34",
  "tracing-subscriber 0.3.11",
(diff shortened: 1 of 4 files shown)


jamesmunns:
No objections from me. It might make sense to do a sit-down review together at some point, and figure out all the ownership/iteration requirements for List? it seems like there's been a couple of conceptual issues that have popped up, it might make sense to make the high level docs/reqs more clear so it's easier to review on this.

PR #210, opened 2022-06-09
Add WaitMap structure

jamesmunns:
Opening this for an early architectural review. Still needs docs and (most) tests updated.

In particular:

I think some more logic regarding "wake_all" needs to be removed: DataWaitQueue doesn't support it, and I think there are some "impossible" control flow items where normally a wake_all could insert an "extra" wake into the queue.

Also, I wonder if I should allow for "failed wakes" - or wakes where NO data is available. This would change the future type to returning `Option<V>`, and would probably be needed to ACTUALLY support `close()` (which now that I think about it, probably panics if you call it now, I haven't re-enabled that test yet)

I also wonder if "WaitQueue" should/could just be a special case of `DataWaitQueue` where the data is `()`, sort of like how I've modified `WaitResult`. That being said, if I do switch to "failed wakes", I don't know if its meaningful to distinguish between `Some(())` and `None`, since a wake is a wake?

Open to thoughts! I need to go off and do other things, but interested in your early feedback on this.


jamesmunns:
Oh also since I wasn't super clear on it: Switching to `Option<V>` would allow a "wake_all" again, by putting `None` in all the completed futures. That being said - I'm not sure if this is useful outside something like the `close` case. The only reason I can think of would be some kind of "hey we're closing, you're never getting your data, sorry".

maitake/src/wait/dataqueue/tests.rs
```diff
@@ -0,0 +1,272 @@
...
+
+    //         let tick = scheduler.tick();
+
+    //         assert_eq!(tick.completed, TASKS);
+    //         assert_eq!(COMPLETED.load(Ordering::SeqCst), TASKS);
+    //         assert!(!tick.has_remaining);
+    //     }
+
+    #[test]
+    fn wake_one() {
```
jamesmunns, line 44: @hawkw I did re-enable this test, which also shows sort of how you'd use this in the "normal" case. This might be worth taking a peek at first.

hawkw:
Some API design thoughts: when I was working on the initial `WaitQueue` implementation, my thought was that we would have something like this:
```rust
pub struct WaitQueue<T = ()> {
    // ...
}

impl<T> WaitQueue<T> {
    pub fn wake_first_match(&self, predicate: impl FnMut(&T) -> bool) {
          // ...
    }

    pub fn wake_all_matches(&self, predicate: impl FnMut(&T) -> bool) {
          // ...
    }

    pub fn wait_with(&self, data: T) -> Wait<'_, T> {
          // ...
    }

}
```
where `T` is arbitrary user-provided data that may be added to a waiter. But, at the time, I decided to punt on that for the initial implementation and add it in a follow-up.

I'm not sure if we actually _want_ the  same user-facing structure to provide both the data-less and data-full wait queue interfaces, but I thought this was worth mentioning.

Another thought, which would take this in a very different direction, is that maybe what we would want is to make this type a `WaitMap` or something, and do away with the queue-like interface entirely. That way, we could replace the linked list with an intrusive red-black tree or something later on.

jamesmunns:
Yeah, I think `WaitMap<K, V>` is a much better name for what I have in mind. In particular, I think the interface already matches that (it only provides a `wake(&K)` interface).

It's not really enforced atm, but the idea is that all `K` instances would be unique. We could handle that one of a few ways:

* Actually check, meaning that `wait()` becomes `O(n)` instead of O(1) - (though it probably stays log(n) if we use a tree), and becomes failable
* Don't check, and state that if you have duplicate `K`s in the map, one of them will be awoken and give the data. we could probably say "first inserted" if it's a queue, it would be harder if the underlying structure is a map.

I'm inclined to do the former and PRETEND we have a tree under the hood (when really it's a list), meaning:

* Today:
    * Insertion is O(1)
    * Wake is O(n)
* With a tree later:
    * Insertion is O(log(n))
    * Wake is O(log(n))

jamesmunns:
Reiterating from chat:

I plan to make the following changes/executive decisions:

* Change the name to WaitMap
* Make `wait(K)` check if the key exists, and panic if it does
* Add a `try_wait(K)` that checks and returns an err if it exists
* Change the wait future to resolve to an `Option<V>`
* Make sure all the data paths still make sense in the absence of a `wake_all` function (though still work with `close()`).

Lemme know if you think of any reason these choices wouldn't be acceptable!

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+    Closed = 0b11,
+}
+
+#[derive(Clone, Debug)]
+enum Wakeup {
+    Empty,
+    Waiting(Waker),
+    One,
+    Closed,
+}
```
hawkw, line 362: it seems like some of these internals could potentially be shared with `WaitQueue`?

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+    }
+}
+
+bitfield! {
+    #[derive(Eq, PartialEq)]
+    struct WaitStateBits<usize> {
+        /// The waiter's state.
+        const STATE: WaitState;
+    }
+}
```
hawkw, line 292: since these are _just_ enums, and don't also contain a count of `wake_all` calls, it seems like we don't need the bitfield and can just convert the `AtomicUsize` directly into a `State` (for the queue), or just store the enum in the waiter directly (for the waiter).

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+        // snapshot the queue's current state.
+        let mut state = self.load();
+
+        // check if any tasks are currently waiting on this queue. if there are
+        // no waiting tasks, store the wakeup to be consumed by the next call to
+        // `wait`.
+        loop {
+            match state.get(QueueState::STATE) {
+                // if the queue is closed, bail.
+                State::Closed => return WakeOutcome::NoMatch(val),
```
hawkw, line 408: is this really `NoMatch`? perhaps there should be an additional `WakeOutcome` variant for "queue is already closed"?

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+    /// impl to be sound.
+    links: list::Links<Waiter<K, V>>,
+
+    /// The node's waker
+    waker: Wakeup,
+
+    // This type is !Unpin due to the heuristic from:
+    // <https://github.com/rust-lang/rust/pull/82834>
+    _pin: PhantomPinned,
+}
```
hawkw, line 267: this is identical to the `Node` in `WaitQueue`...it seems like we could just make this generic over the `Waiter` type, and reuse this code (so the waitmap would have a `Node<Waiter<K, V>>` while `WaitQueue` would have a `Node<Waiter>`).

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+    /// node.
+    ///
+    /// A spinlock (from `mycelium_util`) is used here, in order to support
+    /// `no_std` platforms; when running `loom` tests, a `loom` mutex is used
+    /// instead to simulate the spinlock, because loom doesn't play nice with
+    /// real spinlocks.
+    queue: Mutex<List<Waiter<K, V>>>,
+
+    _key: PhantomData<K>,
+    _val: PhantomData<V>,
```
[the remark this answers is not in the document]
jamesmunns: Correct! Will remove now (and the custom debug impl).

maitake/src/wait/dataqueue.rs
```diff
@@ -0,0 +1,984 @@
...
+            }
+
+            let next = state.with_state(State::Woken);
+            // advance the state to `Woken`, and return (if we did so
+            // successfully)
+            match self.compare_exchange(state, next) {
+                Ok(_) => return WakeOutcome::NoMatch(val),
+                Err(actual) => state = actual,
+            }
+        }
```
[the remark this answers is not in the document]
jamesmunns: Yeah, this is the "data paths that don't make sense". Agreed these should be cleared up (code so far has been pretty "hack and slash").

maitake/src/wait/dataqueue.rs
jamesmunns: No objections! "NoMatch" really means "I didn't wake anyone, here's your data back", but might be worth distinguishing.

maitake/src/wait/dataqueue.rs
jamesmunns: Seems reasonable!

maitake/src/wait/dataqueue.rs
jamesmunns: Yeah, makes sense! I'll address this in the next chunk of work.

maitake/src/wait/dataqueue.rs
```diff
@@ -166,6 +166,7 @@ mod tests;
 /// [ilist]: cordyceps::List
 /// [intrusive]: https://fuchsia.dev/fuchsia-src/development/languages/c-cpp/fbl_containers_guide/introduction
 /// [2]: https://www.1024cores.net/home/lock-free-algorithms/queues/intrusive-mpsc-node-based-queue
+#[derive(Debug)]
```
hawkw, line 169: nit: this debug impl still requires `K` and `V` to be `Debug`, which it maybe shouldn't?

jamesmunns: Does it? Waiter has a custom debug impls, which I think covers it? But I'll make sure before marking this ready.

maitake/src/wait/map.rs
```diff
@@ -458,25 +442,38 @@ impl<K: PartialEq, V> DataWaitQueue<K, V> {
...
+        self.try_wait(key)
+            .expect("Keys should be unique in a `WaitMap`!")
+    }
+
+    pub fn try_wait(&self, key: K) -> Result<Wait<'_, K, V>, ()> {
+        // Check if key already exists
+        let mut mg = self.queue.lock();
+        let mut cursor = mg.cursor();
+        if cursor.find(|n| n.key == key).is_some() {
+            return Err(());
```
hawkw, line 466: do you think this should return the key value?

jamesmunns: Maybe! Probably couldn't hurt.

jamesmunns:
So, a little more progress.

I think I've cleaned up most of the vestigial wake-all state, which should be good.

I also changed my mind again, the Waiter SHOULD resolve to `Result<V>`, not `Result<Option<V>>`, because the only way to wake without data is by closing, and that's an error already.

I *also* am pretty sure I need to add a drop impl somewhere - if we HAVE a V, and the future drops, I'm pretty sure we just drop it without running a dtor, because it's in a MaybeUninit. I'll add a test or three for this.

jamesmunns:
Okay @hawkw, I'd like to call this one ready to merge now.

I believe I have addressed all comments, EXCEPT for refactoring `WaitMap` and `WaitQueue` to have some common internals. If you don't have any argument, I'd like to defer that to a later PR, as I'm a little out of brain energy for this at the moment.

Happy to address any issues or "must fix" review comments!

hawkw:
> I believe I have addressed all comments, EXCEPT for refactoring `WaitMap` and `WaitQueue` to have some common internals. If you don't have any argument, I'd like to defer that to a later PR, as I'm a little out of brain energy for this at the moment.

I'm totally fine with saving that for later!

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+pub enum WaitError {
+    Closed,
+    AlreadyConsumed,
+    NeverAdded,
+}
+
+/// An indication of the result of a call to `wait()`
+// Note: We duplicate the WaitResult type as it currently only has "Closed"
+// as an error condition. We should merge these two on the next breaking
+// change.
```
hawkw, line 43: since this crate has never been published, it's fine to do that now --- although, honestly, i think it's fine (and perhaps better) for the `WaitMap` type to have a different error type than the other types in the wait module. two of those error conditions can never occur with `WaitQueue` or `WaitCell`, so i have a slight preference for them returning an error type where users will not have to match the error and handle two cases that will never happen.

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+use mycelium_bitfield::FromBits;
+use mycelium_util::fmt;
+use mycelium_util::sync::CachePadded;
+use pin_project::{pin_project, pinned_drop};
+
+#[cfg(test)]
+mod tests;
+
+/// An error indicating a failed wake
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
```
hawkw, line 33: this probably should have a 
```suggestion
#[derive(Copy, Clone, Debug, Eq, PartialEq)]
#[non_exhaustive]
```
on it, if there's a possibility of adding new error types? not a blocker.

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+///     assert_eq!(tick.completed, 1);
+/// }
+///
+/// // Tick the scheduler.
+/// let tick = scheduler.tick();
+///
+/// // No additional tasks should be completed
+/// assert_eq!(tick.completed, 0);
+/// assert!(!tick.has_remaining);
+/// ```
```
hawkw, line 127: nitpicky, not a big deal: is there a similar test in the regular `tests` module? i think doctests won't get run with miri etc.

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+            .finish()
+    }
+}
+
+#[pinned_drop]
+impl<K: PartialEq, V> PinnedDrop for Waiter<K, V> {
+    fn drop(self: Pin<&mut Self>) {
+        // SAFETY: If we are being dropped, then we have already been removed
+        // from the map via Waiter::release(). This means we now have exclusive
+        // access to the node, and no lock is required.
```
hawkw, line 243: (or we were never in the map to begin with)

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+
+    /// The Waiter has moved to the `WaitState::Waiting` state. We now
+    /// have the relevant waker, and are still waiting for data. This
+    /// corresponds to `WaitState::Waiting`.
+    Waiting(Waker),
+
+    /// The Waiter has received data, and is waiting for the woken task
+    /// to notice, and take the data by polling+completing the future.
+    /// This corresponds to `WaitState::Completed`.
+    DataReceived,
```
hawkw, line 363: a thought (which is fine to save for a follow-up): i feel like it would make sense to actually store the data in this variant, instead of having a separate `UnsafeCell<MaybeUninit<V>>`? since the `Wakeup` is essentially responsible for determining whether there's a value in there or not. i don't _think_ there's a performance/space improvement from not just storing it here?

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+        test_trace!("wake: -> locked");
+
+        // the queue's state may have changed while we were waiting to acquire
+        // the lock, so we need to acquire a new snapshot.
+        state = self.load();
+
+        if let Some(mut node) = self.node_match_locked(key, &mut *queue, state) {
+            let waker = Waiter::<K, V>::wake(node, &mut *queue, Wakeup::DataReceived);
+            unsafe {
+                node.as_mut().val.with_mut(|v| (*v).write(val));
```
hawkw, line 435: this is safe because we're holding the lock...maybe worth a comment?

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+        }
+    }
+
+    /// Wait to be woken up by this queue.
+    ///
+    /// This returns a [`Wait`] future that will complete when the task is
+    /// woken by a call to [`wake`] with a matching `key`, or when the `WaitMap`
+    /// is dropped.
+    ///
+    /// NOTE: `key`s must be unique. If the given key already exists in the
```
hawkw, line 476: very unimportant nit: usually i bold these
```suggestion
    /// **Note**: `key`s must be unique. If the given key already exists in the
```

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+            .expect("Keys should be unique in a `WaitMap`!")
+    }
+
+    /// Attempt to wait to be woken up by this queue.
+    ///
+    /// This returns a [`Wait`] future that will complete when the task is
+    /// woken by a call to [`wake`] with a matching `key`, or when the `WaitMap`
+    /// is dropped.
+    ///
+    /// NOTE: `key`s must be unique. If the given key already exists in the
```
hawkw, line 494:
```suggestion
    /// **Note**: `key`s must be unique. If the given key already exists in the
```
(but i don't actually care)

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+    /// `WaitMap`, this function will return an error immediately. See [`try_wait`]
+    /// for a non-panicking alternative.
+    ///
+    /// [`try_wait`]: Self::try_wait
+    /// [`wake`]: Self::wake
+    pub fn try_wait(&self, key: K) -> Result<Wait<'_, K, V>, K> {
+        // Check if key already exists
+        let mut mg = self.queue.lock();
```
hawkw, line 502: unimportant style nit: i found the name `mg` kind of un-obvious, but :woman_shrugging:

maitake/src/wait/map.rs
hawkw: (if we stored the data in the `Node`, we could use `Waiter::with_node` to "prove" we're holding the lock...but maybe that's not worth it)

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+
+        opt_node
+    }
+}
+
+pub enum WakeOutcome<V> {
+    Woke,
+    NoMatch(V),
+    Closed(V),
+}
```
hawkw, line 595: these should probably have some quick comments?

maitake/src/wait/map.rs
```diff
@@ -0,0 +1,926 @@
...
+                                retval.assume_init()
+                            });
+
+                            notified(val)
+                        }
+                        Wakeup::Retreived => {
+                            // TODO: Should I change the result type to also return an error
+                            // for this case? Right now error can only be `Closed`
+                            consumed()
+                        }
```
hawkw, line 722: it looks like the TODO comment here was addressed?
```suggestion
                        Wakeup::Retreived => consumed(),
```

maitake/src/wait/map/tests.rs
```diff
@@ -0,0 +1,368 @@
...
+        let scheduler = Scheduler::new();
+        let q = Arc::new(WaitMap::<CountDropKey, CountDropVal>::new());
+
+        const TASKS: usize = 10;
+
+        for i in 0..TASKS {
+            let q = q.clone();
+            scheduler.spawn(async move {
+                // TODO(AJM): I need to select!() against a one shot channel or something
+                // for the "wait hanging" test
```
hawkw, line 178: did you want to address this before merging?

maitake/src/wait/map/tests.rs
```diff
@@ -0,0 +1,368 @@
...
+        assert!(!tick.has_remaining);
+
+        q.close();
+
+        let tick = scheduler.tick();
+
+        assert_eq!(tick.completed, TASKS);
+        assert_eq!(COMPLETED.load(Ordering::SeqCst), TASKS);
+        assert_eq!(KEY_DROPS.load(Ordering::SeqCst), TASKS);
+        assert_eq!(VAL_DROPS.load(Ordering::SeqCst), 0);
```
hawkw, line 157: nit, take it or leave it: all the atomic ops in these tests *can* be `Relaxed` --- the tests are all single-threaded, so these atomics will never be modified anywhere else, so ordering doesn't matter at all. also, they're just counters, so they can always be `Relaxed`. not a big deal though.

hawkw (approved):
looks good to me overall! i left a few minor suggestions.

maitake/src/wait/map.rs
jamesmunns: Makes sense! Will remove the comment.

maitake/src/wait/map.rs
jamesmunns: Yup! This is basically a duplicate of `maitake::wait::map::tests::wake_one()`.

maitake/src/wait/map.rs
jamesmunns: That's a good point! I've added a TODO comment.

maitake/src/wait/map.rs
jamesmunns: Reasonable! Switched to `guard`.

maitake/src/wait/map/tests.rs
jamesmunns: Good catch! I already added the `drop_wake_bailed()` test, which addresses this.

jamesmunns:
Addressed all comments, marking this for auto-merge if CI passed.

PR #211, opened 2022-06-09
feat(cordyceps): add `List::{front, back, front_mut, back_mut}`

hawkw:
This branch adds accessors for the front and back elements of a
`cordyceps::List`. The returned elements are `Pin`ned to prevent them
from being moved out of the list. The `Pin`ning was not strictly
necessary for the immutable accessors but I did it for consistency with
the mutable ones...we _could_ safely just have those return `&T`,
though.


jamesmunns (approved):
Overall LGTM, up to you if you want to add (regular/miri/loom) tests here.

PR #212, opened 2022-06-09
feat(cordyceps): add `List::drain_filter`

hawkw:
This branch adds a `List::drain_filter` iterator that moves items out of
the list if they match a predicate. This is based on the
[`drain_filter`] iterator on the `std::collections::LinkedList` type,
but modified slightly.

[`drain_filter`]: https://doc.rust-lang.org/stable/std/collections/struct.LinkedList.html#method.drain_filter

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -264,6 +264,19 @@ pub struct IterMut<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
+/// An iterator returned by [`List::drain_filter`].
+pub struct DrainFilter<'list, T, F>
+where
+    F: FnMut(&T) -> bool,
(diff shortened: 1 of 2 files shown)


cordyceps/src/list.rs
```diff
@@ -264,6 +264,19 @@ pub struct IterMut<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
+/// An iterator returned by [`List::drain_filter`].
+pub struct DrainFilter<'list, T, F>
```
jamesmunns, line 268: Would it make sense to re-use `Cursor` here so we don't duplicate the iteration logic? You should be able to re-borrow the cursor mutable view into an immutable view for the predicate.

jamesmunns (approved):
Overall LGTM, one note about code re-use.

PR #216, opened 2022-06-12
A second round of additions to wait::WaitMap.

jamesmunns:
This includes the following changes:

Moves the uniqueness check from future creation point to
the time of enqueuing - otherwise you could have multiple
futures created with the same key, but not enqueued, that
would all be enqueued on their first poll. This also
removes the "try_wait" function, and "wait" can no longer
panic. In the duplicate case, the future will resolve to
an error instead.

Adds a EnqueueWait future that allows users to await ONLY
until the Wait has been enqueued in the actual WaitMap,
meaning they can then perform other operations without
a risk of racing with the expected response.

Corrected some incorrect documentation from the initial
Pull Request

Added tests for the above described behaviors

### DIFF
--- a/maitake/src/wait/map.rs
+++ b/maitake/src/wait/map.rs
@@ -33,16 +33,18 @@ mod tests;
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 #[non_exhaustive]
 pub enum WaitError {
-    /// The WaitMap has already been closed
+    /// The [`WaitMap`] has already been [closed].
+    ///
+    /// [closed]: WaitMap::close
     Closed,
 
     /// The received data has already been extracted
     AlreadyConsumed,
 
-    /// The Wait was never added to the WaitMap
+    /// The [`Wait`] was never added to the [`WaitMap`]
     NeverAdded,
 
-    /// The WaitMap already had an item matching the given
+    /// The [`WaitMap`] already had an item matching the given
     /// key
     Duplicate,
 }
@@ -218,7 +220,7 @@ pub struct Wait<'a, K: PartialEq, V> {
 
 impl<'map, 'wait, K: PartialEq, V> Wait<'map, K, V> {
     /// Returns a future that completes when the `Wait` item has been
-    /// added to the `WaitMap`, and is ready to receive data
+    /// added to the [`WaitMap`], and is ready to receive data
     ///
     /// This is useful for ensuring that a receiver is ready before
     /// sending a message that will elicit the expected response.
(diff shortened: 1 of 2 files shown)


maitake/src/wait/map.rs
```diff
@@ -33,9 +33,18 @@ mod tests;
...
     Closed,
+
+    /// The received data has already been extracted
     AlreadyConsumed,
+
+    /// The Wait was never added to the WaitMap
     NeverAdded,
+
+    /// The WaitMap already had an item matching the given
+    /// key
```
hawkw, line 46: tiny nit:
```suggestion
    /// The [`WaitMap`] has already been [closed].
    ///
    /// [closed]: WaitMap::close
    Closed,

    /// The received data has already been extracted
    AlreadyConsumed,

    /// The [`Wait`] was never added to the [`WaitMap`]
    NeverAdded,

    /// The [`WaitMap`] already had an item matching the given
    /// key
```

maitake/src/wait/map.rs
```diff
@@ -204,6 +216,56 @@ pub struct Wait<'a, K: PartialEq, V> {
     waiter: Waiter<K, V>,
 }
 
+impl<'map, 'wait, K: PartialEq, V> Wait<'map, K, V> {
+    /// Returns a future that completes when the `Wait` item has been
+    /// added to the `WaitMap`, and is ready to receive data
```
hawkw, line 221: nit: maybe
```suggestion
    /// added to the [`WaitMap`], and is ready to receive data
```

maitake/src/wait/map.rs
```diff
@@ -607,6 +634,27 @@ pub enum WakeOutcome<V> {
 
 // === impl Waiter ===
 
+/// A future that ensures a [`Wait`] has been added to the [`WaitQueue`].
```
hawkw, line 637: nit: ```suggestion
/// A future that ensures a [`Wait`] has been added to a [`WaitMap`].
```

Comment (hawkw) [inline maitake/src/wait/map.rs:637]: this should probably not be under `// === impl Waiter ===` :)

File: maitake/src/wait/map.rs [hunk 4]
```diff
@@ -204,6 +216,56 @@ pub struct Wait<'a, K: PartialEq, V> {
...
+    ///     assert_eq!(val, 10);
+    /// });
+    ///
+    /// assert!(matches!(q.wake(&0, 100), WakeOutcome::NoMatch(_)));
+    ///
+    /// let tick = scheduler.tick();
+    ///
+    /// assert!(matches!(q.wake(&0, 100), WakeOutcome::Woke));
+    /// ```
+    pub fn enqueue(self: Pin<&'wait mut Self>) -> EnqueueWait<'wait, 'map, K, V> {
```
hawkw, line 264: API-wise, it feels like kind of a bummer that this doesn't take the `Wait` by value and then return it, but it has to be pinned to be added to the queue, and can't move after it's enqueued...so we couldn't implement an API like that without boxing. so, this seems good :+1:

hawkw (approved):
overall, this looks great to me! i commented on a handful of documentation nits, but none of them really matter.

PR #218, opened 2022-06-13
Raise spawn requirements to require + 'static

### LINKED ISSUE #217: maitake: spawn_* methods should probably enforce `F: Future + 'static`
I think at the moment, the spawn and spawn_allocated functions on the scheduler are too permissive. You could (maybe?) spawn a future that refers to a local, then hold on to it for longer than necessary.

jamesmunns:
Closes #217

### DIFF
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -82,7 +82,7 @@ impl StaticScheduler {
     #[inline]
     pub fn spawn_allocated<F, STO>(&'static self, task: STO::StoredTask)
     where
-        F: Future,
+        F: Future + 'static,
         STO: Storage<&'static Self, F>,
     {
         let tr = TaskRef::new_allocated::<&'static Self, F, STO>(task);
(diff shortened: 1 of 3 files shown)


jamesmunns:
Looks like the test failures are due to other places now requiring `async move` instead of `async`.

As an example of why this might be Bad, here's a test:

```rust
    #[test]
    fn notify_future() {
        static STUB: TaskStub = TaskStub::new();
        static SCHEDULER: StaticScheduler = unsafe { StaticScheduler::new_with_static_stub(&STUB) };
        static COMPLETED: AtomicUsize = AtomicUsize::new(0);

        crate::util::trace_init();
        let chan = Chan::new(1);

        MyBoxTask::spawn(&SCHEDULER, {
            let chan = chan.clone();
            async move {
                chan.wait().await;
                COMPLETED.fetch_add(1, Ordering::SeqCst);
            }
        });

        MyBoxTask::spawn(&SCHEDULER, async {
            Yield::once().await;
            chan.notify();
        });

        dbg!(SCHEDULER.tick());

        assert_eq!(COMPLETED.load(Ordering::SeqCst), 1);
    }
```

If I were to change it to this:

```rust
    #[test]
    fn notify_future() {
        static STUB: TaskStub = TaskStub::new();
        static SCHEDULER: StaticScheduler = unsafe { StaticScheduler::new_with_static_stub(&STUB) };
        static COMPLETED: AtomicUsize = AtomicUsize::new(0);

        crate::util::trace_init();
        let chan = Chan::new(1);

        MyBoxTask::spawn(&SCHEDULER, {
            let chan = chan.clone();
            async move {
                chan.wait().await;
                COMPLETED.fetch_add(1, Ordering::SeqCst);
            }
        });

        MyBoxTask::spawn(&SCHEDULER, async {
            Yield::once().await;
            chan.notify();
        });

        drop(chan); // This sure looks like I drop a local that is held...

        dbg!(SCHEDULER.tick());

        assert_eq!(COMPLETED.load(Ordering::SeqCst), 1);
    }
```

The second example STILL COMPILES, which I think is Bad.

hawkw:
Thanks, that was a (pretty bad) oversight on my part. Good catch!

PR #219, opened 2022-06-15
feat(maitake): add an async `Mutex`

hawkw:
This branch adds an async `Mutex` type to `maitake`, implemented using
`WaitQueue`. The implementation should be pretty straightforward,
although it required adding some additional methods to `WaitQueue`.
Currently, all the new `WaitQueue` functions are `pub(crate)`, but I'm
open to making them public APIs as well; I just wanted to save that for
a separate PR.

The `Mutex` type lives in a new `maitake::sync` module for
synchronization primitives --- I was on the fence about whether it still
makes sense to have a separate `maitake::wait` module if we are going to
have `maitake::sync`, but I didn't want to put the mutex in `wait`, as
it does more than *just* notify waiting tasks. Perhaps the `wait` types
should be moved into `sync`, but I'd prefer to do that in a separate PR.

### DIFF
--- a/maitake/src/sync/mutex.rs
+++ b/maitake/src/sync/mutex.rs
@@ -134,7 +134,7 @@ impl<'a, T> Deref for MutexGuard<'a, T> {
     }
 }
 
-impl<'a, T> DerefMut for MutexGuard<'a, T> {
+impl<T> DerefMut for MutexGuard<'_, T> {
     #[inline]
     fn deref_mut(&mut self) -> &mut Self::Target {
         unsafe {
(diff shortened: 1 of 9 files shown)


maitake/src/sync/mutex.rs
```diff
@@ -0,0 +1,487 @@
...
+///
+/// [^1]: And therefore require an operating system to manage threading.
+///
+/// [^2]: The [tasks][crate::task::Task] themselves must, of course, be stored
+///     somewhere, but this need not be a heap allocation in systems with a
+///     fixed set of  statically-allocated tasks. And, when tasks *are*
+///     heap-allocated, these  allocations [need not be provided by
+///     `liballoc`][storage].
+///
+/// [^3]: In fact, this mutex _cannot_ implement poisoning, as poisoning
```
jamesmunns, line 62: Is this documented anywhere else?

maitake/src/sync/mutex.rs
```diff
@@ -0,0 +1,487 @@
...
+            _wake: WakeOnDrop(self),
+            data: self.data.get_mut(),
+        }
+    }
+}
+
+impl<T: ?Sized + fmt::Debug> fmt::Debug for Mutex<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("Mutex")
+            .field("data", &fmt::opt(&self.try_lock()).or_else("<locked>"))
```
jamesmunns, line 238: :100:

jamesmunns (approved):
Overall lgtm!

maitake/src/sync/mutex.rs
hawkw: no, i keep meaning to add it to the crate level docs

jamesmunns: Not a blocker! I just wanted to cross link it if there was a larger discussion somewhere.

PR #221, opened 2022-06-16
Remove unused mycotest dep

jamesmunns:
This removes an unused `mycotest` dependency in maitake.

This was causing `#![no_std]` builds to fail as `mycotest` pulls in `once_cell`, which isn't marked as `no_std` (at least in the used feature configuration).

We should probably add tests in CI to all of the `no_std` crates to make sure that a `cargo build --no-default-features --target thumbv7em-none-eabihf` runs successfully to avoid regressions.

### DIFF
--- a/maitake/Cargo.toml
+++ b/maitake/Cargo.toml
@@ -36,7 +36,6 @@ no-cache-pad = ["mycelium-util/no-cache-pad", "cordyceps/no-cache-pad"]
 [dependencies]
 mycelium-bitfield = { path = "../bitfield" }
 mycelium-util = { path = "../util" }
-mycotest = { path = "../mycotest", default-features = false }
 cordyceps = { path = "../cordyceps" }
 pin-project = "1"
 


hawkw (approved):
Thanks, good catch --- there used to be mycotests in `maitake` but I moved them to the kernel because I didn't want them showing up when non-mycelium projects depend on maitake: https://github.com/hawkw/mycelium/blob/bb21d17ef5fed748786d71d983db515027fe943e/src/tests.rs#L42-L148

The `default-features = false` version of `mycotest` _shouldn't_ pull anything that doesn't build on `no-std`, though, as it's intended to be used in mycelium crates on `no-std`...so that's weird. I might have to look into that separately.

hawkw:
Than

> The `default-features = false` version of `mycotest` _shouldn't_ pull anything that doesn't build on `no-std`, though, as it's intended to be used in mycelium crates on `no-std`...so that's weird. I might have to look into that separately.

Oh, I think the issue is related to `mycotest` depending on `tracing`, and there being a patch in the workspace to pull `tracing` 0.2 from Git, but when that patch isn't present, maybe it pulls an older tracing, or something?

PR #223, opened 2022-06-16
`wait` function for `WaitCell` and also rework WaitCell logic

jamesmunns:


maitake/src/wait/cell.rs
```diff
@@ -56,26 +91,24 @@ impl WaitCell {
 }
 
 impl WaitCell {
-    pub fn poll_wait(&self, waker: &Waker) -> Poll<WaitResult<()>> {
+    fn register_wait(&self, waker: &Waker) -> Result<(), Error> {
```
jamesmunns, line 94: Updated the name and signature.

The name makes more sense (to me), since you aren't really polling for the wait to be ready. You're attempting to register a new waiter to the Wait Cell.

The type also changed, because you CANNOT return `Poll::Ready(Ok(()))` here anymore, only an error or a pending.

`Ok(())` means you registered successfully, which maps to `Poll::Pending`.

maitake/src/wait/cell.rs
```diff
@@ -56,26 +91,24 @@ impl WaitCell {
...
             // someone else is notifying, so don't wait!
             Err(actual) if test_dbg!(actual.is(State::CLOSED)) => {
-                return wait::closed();
+                return closed();
             }
             Err(actual) if test_dbg!(actual.is(State::NOTIFYING)) => {
-                waker.wake_by_ref();
-                crate::loom::hint::spin_loop();
-                return wait::notified(());
+                return busy();
```
jamesmunns, line 104: Here we declare that the waitcell is busy, and it is now the callers responsibility to do with that as they will.

Though now that I've made this function private, it means that `Wait` will always pend it's own waker and try again. That might be the wrong behavior, or something we configure in the call to `wake`.

maitake/src/wait/cell.rs
```diff
@@ -56,26 +91,24 @@ impl WaitCell {
...
             }
 
             Err(actual) => {
                 debug_assert!(
                     actual == State::PARKING || actual == State::PARKING | State::NOTIFYING
                 );
-                return Poll::Pending;
+                return busy();
```
jamesmunns, line 111: If we're in the middle of some other operation, we can't possibly have added this new waker to the cell. Report it as busy.

maitake/src/wait/cell.rs
```diff
@@ -88,8 +121,13 @@ impl WaitCell {
             }
         });
 
+        if let Some(prev_waker) = prev_waker {
```
jamesmunns, line 124: Before we would only wake the "old" waker in the cell in the case of some kind of spurious close event.

Now, if you replace one waker with another, the old one gets woken as if the wait completed. This could lead to two wakers fighting over the cell, but hey, whatcha gunna do, don't use a WaitCell.

maitake/src/wait/cell.rs
```diff
@@ -103,23 +141,25 @@ impl WaitCell {
...
 
-                if let Some(prev_waker) = prev_waker {
-                    prev_waker.wake();
-                }
-
                 if let Some(waker) = waker {
                     waker.wake();
                 }
 
-                if test_dbg!(state.is(State::CLOSED)) {
```
jamesmunns, line 114: I don't *think* this logic can ever happen with the given states, and returning "notified" didn't make sense to me. Let me know if I screwed some logic up.

maitake/src/wait/cell.rs
```diff
@@ -189,6 +229,32 @@ impl fmt::Debug for WaitCell {
...
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        if !self.registered {
+            match self.cell.register_wait(cx.waker()) {
+                Ok(_) => {
+                    self.registered = true;
+                    Poll::Pending
+                }
+                Err(Error::Busy) => {
+                    // Cell was busy, all we can do is try again later
```
jamesmunns, line 245: This is where we could maybe have a configurable whether "busy" means "okay cool waited we're done here", or "keep trying". I think the as-is behavior is a reasonable position though.

maitake/src/wait/cell.rs
```diff
@@ -271,17 +367,14 @@ pub(crate) mod test_util {
                     None => return Poll::Ready(()),
                 };
 
-                let res = test_dbg!(this.task.poll_wait(cx.waker()));
+                let res = this.task.wait();
```
jamesmunns, line 370: I admit to not totally understanding this test, but I've more or less done a 1:1 port with the new, only-`wait()` API.

jamesmunns:
I've broken the loom tests:

```
running 2 tests
test wait::cell::loom::basic_latch ... thread 'main' panicked at 'deadlock; threads = [(Id(0), Blocked(Location(Some(Location { file: "maitake/src/wait/cell.rs", line: 418, col: 13 })))), (Id(1), Terminated), (Id(2), Terminated)]', /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/loom-0.5.5/src/rt/execution.rs:215:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /home/james/.cargo/registry/src/github.com-1ecc6299db9ec823/loom-0.5.5/src/rt/thread.rs:259:39
```

Not what's up, will attempt to diagnose later, but I've probably done a dumb.

maitake/src/wait/cell.rs
hawkw: i think the name and signature changes make sense here, the `poll_wait` design was me trying to be too clever, and i think maybe it never made that much sense in most cases.

i *do* think, however, this method probably should still be `pub`. 

i can imagine wanting to register a waiter outside of a `Wait` future (for example, in a `JoinHandle`). in particular, the `Wait` future borrows the `WaitCell`, and i think a common use case might involve registering a waker with a `WaitCell` in something that doesn't have a non-`'static` lifetime --- for example, the `WaitCell` might be inside of some other shared structure that's `Arc`ed, and we might want to clone that `Arc` into the future that registers, or something... 

```suggestion
    pub fn register_wait(&self, waker: &Waker) -> Result<(), Error> {
```

maitake/src/wait/cell.rs
```diff
@@ -189,6 +229,32 @@ impl fmt::Debug for WaitCell {
...
+                    cx.waker().wake_by_ref();
+                    Poll::Pending
+                }
+                Err(Error::Closed) => Poll::Ready(Err(Error::Closed)),
+            }
+        } else {
+            // We made it to "once", and got polled again, We must be ready!
+            Poll::Ready(Ok(()))
+        }
+    }
```
hawkw, line 255: very tiny style nit, take it or leave it: i think it would be a little bit simpler to write this instead:
```suggestion
        if self.registered {
            // We made it to "once", and got polled again, We must be ready!
            return Poll::Ready(Ok(()));
        }
        
        match self.cell.register_wait(cx.waker()) {
            Ok(_) => {
                self.registered = true;
                Poll::Pending
            }
            Err(Error::Busy) => {
                // Cell was busy, all we can do is try again later
                cx.waker().wake_by_ref();
                Poll::Pending
            }
            Err(Error::Closed) => Poll::Ready(Err(Error::Closed)),
        }
    }
```

maitake/src/wait/cell.rs
hawkw: yeah, i think there could be a `wait_retrying` method (or something), and a bool on the future to configure the behavior. but, i think for now, let's not add stuff to the API unless we *need* to use it for something --- the current approach seems fine and we can always go back and make it configurable later.

maitake/src/wait/cell.rs
```diff
@@ -103,23 +141,25 @@ impl WaitCell {
...
-                    wait::notified(())
-                }
+                // We just went to the closed state, so sorry new waker, shops
+                // closed
+                closed()
             }
         }
     }
 
+    /// Wait to be woken up by this cell.
```
hawkw, line 155: an example could be nice, but we can add that later.

it might also be nice to explicitly state that this does _not_ register the waker until the first time the future is polled...

maitake/src/wait/cell.rs
```diff
@@ -7,14 +7,35 @@ use crate::{
...
+    future::Future,
     ops,
-    task::{Poll, Waker},
+    pin::Pin,
+    task::{Context, Poll, Waker},
 };
 use mycelium_util::{fmt, sync::CachePadded};
 
+/// An error indicating that a [`WaitCell`] or [`WaitQueue`] was closed while
+/// attempting register a waiter.
```
hawkw, line 20: i think this doc comment needs to be updated --- this is only returned by `WaitCell`, and can indicate that the cell is closed _or_ busy.

maitake/src/wait/cell.rs
```diff
@@ -7,14 +7,35 @@ use crate::{
...
+    Err(Error::Closed)
+}
+
+pub const fn registered() -> Result<(), Error> {
+    Ok(())
+}
+
+pub const fn busy() -> Result<(), Error> {
+    Err(Error::Busy)
+}
```
hawkw, line 37: do these need to be public?
```suggestion
const fn closed() -> Result<(), Error> {
    Err(Error::Closed)
}

const fn registered() -> Result<(), Error> {
    Ok(())
}

const fn busy() -> Result<(), Error> {
    Err(Error::Busy)
}
```

maitake/src/wait/cell.rs
```diff
@@ -189,6 +229,32 @@ impl fmt::Debug for WaitCell {
     }
 }
 
+// === impl Wait ===
+
+impl Future for Wait<'_> {
+    type Output = Result<(), Error>;
```
hawkw, line 235: this never returns `Error::Busy`...I think it should probably be changed to return the `wait::Closed` error type, rather than `Error`.

if we're going to add a variant of this future that behaves differently when the cell is busy, I think it would make sense for that future to have a different error type.

maitake/src/wait/cell.rs
```diff
@@ -103,23 +141,25 @@ impl WaitCell {
...
 
+    /// Wait to be woken up by this cell.
+    pub fn wait(&self) -> Wait<'_> {
+        Wait {
+            cell: self,
+            registered: false,
+        }
+    }
+
     pub fn notify(&self) -> bool {
```
hawkw, line 163: this is completely unrelated to this change, but i think we should probably rename this method to `wake`, for consistency with the other primitives in `mycelium_util::wake`. we can do that in a separate PR, but it was just something i noticed while i was here.

hawkw (commented):
overall, i think this change looks correct --- it's actually making `WaitCell` closer to [the `AtomicWaker` type in `futures`](https://docs.rs/futures/latest/futures/task/struct.AtomicWaker.html) which i (perhaps foolishly) thought i could improve on the interface of with the `poll_wait` method.

i left a bunch of minor style comments, but i'm +1 on the actual change here.

jamesmunns:
@hawkw this should be ready to merge now. I have performed some test crimes, but I think they are necessary. I am open to better solutions, but this is one that works.

jamesmunns:
Okay, I've nailed down the loom failures to basically the fact that a WaitCell can't be guaranteed to terminate if two threads are involved, as loom will sometimes decide to totally starve one thread, and then timeout the other.

I've "fixed" one test (by bailing after N attempts), and ignored another. Open to other suggestions.

maitake/src/wait/cell.rs
```diff
@@ -237,8 +303,37 @@ impl fmt::Debug for State {
     }
 }
 
+#[cfg(all(feature = "alloc", not(loom), test))]
+mod test {
```
hawkw, line 307: this is a _very_ stupid nit, and i'm sorry in advance, but...every other module in this crate calls the test module `tests` rather than `test`, and it's kind of nice to be consistent with that when running with `cargo test wait::cell::tests` or similar...

maitake/src/wait/cell.rs
```diff
@@ -237,8 +303,37 @@ impl fmt::Debug for State {
...
+        let tick = sched.tick();
+        assert_eq!(tick.completed, 0);
+        assert_eq!(COMPLETED.load(Ordering::Relaxed), 0);
+
+        assert_eq!(wait.wake(), true);
+        let tick = sched.tick();
+        assert_eq!(tick.completed, 1);
+        assert_eq!(COMPLETED.load(Ordering::Relaxed), 1);
+    }
+}
```
hawkw, line 334: also, it occurs to me that since this is testing only `WaitCell`'s public API, it _might_ belong in `tests/` as an integration test, rather than in `wait::cell::tests` as a unit test...but...i don't actually care...

maitake/src/wait/cell.rs
```diff
@@ -102,24 +138,28 @@ impl WaitCell {
...
+                // closed
+                closed()
             }
         }
     }
 
-    pub fn notify(&self) -> bool {
+    /// Wait to be woken up by this cell.
+    ///
+    /// Note: The waiter is not registered until AFTER the first time the waiter is polled
```
hawkw, line 154: i wonder if it would make sense to add something like `WaitMap::enqueue` here, for eagerly registering a `Wait` future? but we can add that in a follow-up PR.

maitake/src/wait/cell.rs
```diff
@@ -188,6 +228,32 @@ impl fmt::Debug for WaitCell {
...
 }
 
+// === impl Wait ===
+
+impl Future for Wait<'_> {
+    type Output = Result<(), super::Closed>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        if self.registered {
+            // We made it to "once", and got polled again, We must be ready!
```
hawkw, line 238: typo (sorry):
```suggestion
            // We made it to "once", and got polled again, we must be ready!
```

maitake/src/wait/cell.rs
```diff
@@ -306,45 +399,63 @@ pub(crate) mod test_util {
...
 
-            thread::spawn({
-                let chan = chan.clone();
-                move || {
-                    chan.notify();
-                }
+            thread::spawn(move || {
+                info!("waking");
+                waiter.wake();
+                info!("woke'd");
```
hawkw, line 416: extremely silly, sorry: there _is_ actually a word that means "woke'd", it's "woken"...but this is funny, so i am fine with leaving it.

maitake/src/wait/cell.rs
hawkw: i think the problem here is that we have two different "busy" states, the state where another thread is parking a waiter, and the state where the cell is being woken. i don't actually think it's correct to return `busy` in the case where the cell is being woken. in the previous impl, this would wake the waiter, and return `Ready`...i'm not sure if it's correct for the `wait` future to busy-loop forever if it tries to register while the cell is in the middle of waking, i think that should just complete the future, instead?

maitake/src/wait/cell.rs
```diff
@@ -306,45 +399,63 @@ pub(crate) mod test_util {
...
-            thread::spawn({
-                let chan = chan.clone();
-                move || {
-                    chan.close();
+            info!("waiting");
+            let _ = future::block_on(async move {
+                // Unfortunately, SOMETIMES loom just wants to preempt the 'waker' task
+                // JUST as it gains access to the lock, but BEFORE actually waking the pending task,
+                // never yielding back to the waker task, which then just essentially leaves it stuck
+                // in the NOTIFYING state.
```
hawkw, line 429: this...feels like kind of a serious bug, to me? i left some comments on the `register_wait` function about what i _think_ we should do here, but it's possible i've forgotten some of the context...

maitake/src/wait/cell.rs
jamesmunns: Agree, but will do in a follow up PR, kind of spent on this one ATM, and would like to get it merged.

hawkw (approved):
yay, this looks good to me! thank you!

PR #228, opened 2022-06-19
More Miri Checks!

jamesmunns:
Enable strict provenance checks from https://gankra.github.io/blah/tower-of-weakenings/#ok-but-what-the-heck-is-strict-provenance

### DIFF
--- a/bin/miri
+++ b/bin/miri
@@ -6,6 +6,6 @@ set -x
 
 cd "$(dirname "$0")"/..
 
-MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tag-raw-pointers ${MIRIFLAGS:-}" \
+MIRIFLAGS="-Zmiri-disable-isolation -Zmiri-tag-raw-pointers -Zmiri-strict-provenance ${MIRIFLAGS:-}" \
     PROPTEST_CASES="${PROPTEST_CASES:-10}" \
     cargo miri test --lib "${@}"


hawkw:
I wonder if we ought to change the CI configuration so that instead of running the `bin/miri` script which enables all miri checks at once, it runs separate jobs with different miri flags enabled? That way, we can run the different checks in parallel instead of enabling them all in the same run.

My understanding is that these flags only control checking, and don't change the _behavior_ of the interpreter (unlike other Miri flags like the one that randomizes struct layout), so running the different checks in separate jobs _shouldn't_ result in less exhaustive checking AFAIK?

hawkw:
...huh, it claims that the Cordyceps miri run finished in 2 mins on CI, so maybe it's not a big deal and we can enable all the flags at once: https://github.com/hawkw/mycelium/runs/6951525208?check_suite_focus=true

hawkw (approved):
looking at rust-lang/miri#2045, it seems like `-Zmiri-strict-provenance` actually _implies_ `-Zmiri-tag-raw-pointers`, so they're not actually disjoint sets of checking at all (and we _could_ remove the `tag-raw-pointers` flag, although maybe it's nice to leave it on explicitly?).

we may also want to enable `-Zmiri-check-number-validity` and `-Zmiri-symbolic-alignment-check` to our Miri script, but we can do that in a separate PR.

let's go ahead and merge this one!

PR #248, opened 2022-07-07
feat(maitake): add feature flags for both versions of `tracing`

hawkw:
This branch adds support for emitting both `tracing` 0.1 and `tracing`
0.2 diagnostics from Maitake. They are both controlled by feature flags,
which are additive --- both `tracing` versions _can_ be supported
simultaneously. The `tracing` configuration for tests still works as it
did previously.

The `tracing` 0.2 feature flag probably can't stay forever, if we ever
want to do a crates.io release of `maitake` before `tracing` 0.2 is
done...

### DIFF
--- a/maitake/README.md
+++ b/maitake/README.md
@@ -132,10 +132,13 @@ The following features are available (this list is incomplete; you can help by [
 
 [expanding it]: https://github.com/hawkw/mycelium/edit/main/maitake/README.md
 
-| Feature | Default | Explanation |
-| :---    | :---    | :---        |
-| `alloc` | `true`  | Enables [`liballoc`] dependency |
+| Feature        | Default | Explanation |
+| :---           | :---    | :---        |
+| `alloc`        | `true`  | Enables [`liballoc`] dependency |
 | `no-cache-pad` | `false` | Inhibits cache padding for the [`CachePadded`] struct. When this feature is NOT enabled, the size will be determined based on target platform. |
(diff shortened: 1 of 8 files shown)


maitake/README.md
```diff
@@ -132,10 +132,13 @@ The following features are available (this list is incomplete; you can help by [
...
 [expanding it]: https://github.com/hawkw/mycelium/edit/main/maitake/README.md
 
-| Feature | Default | Explanation |
-| :---    | :---    | :---        |
-| `alloc` | `true`  | Enables [`liballoc`] dependency |
+| Feature        | Default | Explanation |
+| :---           | :---    | :---        |
+| `alloc`        | `true`  | Enables [`liballoc`] dependency |
 | `no-cache-pad` | `false` | Inhibits cache padding for the [`CachePadded`] struct. When this feature is NOT enabled, the size will be determined based on target platform. |
+| `tracing-01`   | `false`  | Enables support for v0.1.x of [`tracing`] (the current release version). Requires `liballoc`.|
```
jamesmunns, line 139: Wait, does tracing 0.1 still require `liballoc`? I thought we had squished that away. I was able to build a no-std lib for thumbv7em-none-eabihf with tracing v0.1.35.

hawkw: Tracing 0.1 will always require liballoc, it supports no-std + alloc. Getting rid of alloc in 0.1 isn't possible as it would require feature-flagging APIs that are currently available on `default-features = false`.

PR #250, opened 2022-07-09
feat(maitake): add a task builder for naming tasks

hawkw:
This branch adds a `task::Builder` API to `maitake` for configuring task
settings prior to spawning. Currently, the only settings that can be
configured are metadata fields used to populate the `tracing` span
generated for that task, such as a name (used for the `tokio-console`
`task.name` field).

This is similar to the (unstable) `task::Builder` API in `tokio`.


jamesmunns (approved):
Looks great!

Issue #238, opened 2022-06-23
maitake: scheduler trait is kind of weird


hawkw:
Currently, `maitake` has a `Schedule` trait that abstracts over a scheduler that can schedule a task:
https://github.com/hawkw/mycelium/blob/653aaf8d6040ad4146b0330ef957e26ae904a7b6/maitake/src/scheduler.rs#L27

This has to be a public trait, as it's one of the trait bounds on `Task`'s generics. However, the trait is kind of weird if user code intends to implement it. The only way to schedule a task is to push it to the scheduler's run queue, which is a `cordyceps::MpscQueue`. However, the `Linked<mpsc_queue::Links<...>>` impl for `task::Header` is private, so downstream code cannot use this to implement `Schedule` for its own types.

There are some questions we may eventually want to work out:

1. do we want `Schedule` to be an extension point? is there a reason for user code to implement `Schedule` for its own types?
2. if we want user code to implement this trait, should we consider changing it from having a `schedule` method to having a `run_queue` accessor or similar? that way, the `Task` code can handle actually pushing itself to the scheduler run queue, and it just asks that a scheduler own a run queue.
3. if we *don't* want user code to implement `Schedule`, should we 
  - remove the trait and replace it with some kind of type erased `SchedulerRef` type (potentially adding dyn dispatch overhead...)?
  - make the trait impossible to implement in user code *explicitly*?
  - just document that this isn't intended to be implemented downstream?

hawkw:
cc @jamesmunns any opinions?

jamesmunns:
So, "as a downstream user", I don't know if I'd necessarily want to be able to abstract over the scheduler that I'm working with? But if implementing the trait requires intimate knowledge of the scheduler implementation like you've said above, I'd probably say it's better to:

* Make it a sealed trait (I did this in #251, mnemos still builds with this change)
* Remove the trait entirely

It's acceptable to make it a sealed trait, as the user MUST choose one of the Schedule impls provided by Maitake, and can use them to fulfill the `S: Schedule` trait bound, though they can't use that to make THIER downstream types generic over some S in the future (because they can't name the Schedule trait).

This seems fine to me? But again, I'm not trying to be generic over my scheduler implementations. If you can think of a reason that someone would want to do that, you could close #251, but if not, I'd say merge it, or just yeet the trait entirely.

PR #251, opened 2022-07-11
Make Schedule a sealed trait

jamesmunns:
Open as a discussion point to #238

### DIFF
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -21,8 +21,13 @@ pub struct Tick {
     pub has_remaining: bool,
 }
 
-pub trait Schedule: Sized + Clone {
-    fn schedule(&self, task: TaskRef);
+use sealed::Schedule as _;
+pub(crate) mod sealed {
+    use crate::task::TaskRef;
+
+    pub trait Schedule: Sized + Clone {
+        fn schedule(&self, task: TaskRef);
+    }
 }
 
 /// A stub [`Task`],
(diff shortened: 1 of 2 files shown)


maitake/src/scheduler.rs
```diff
@@ -21,16 +21,22 @@ pub struct Tick {
...
 
-    /// Returns a new [task `Builder`] for configuring tasks prior to spawning
-    /// them on this scheduler.
-    ///
-    /// [task `Builder`]: task::Builder
-    #[must_use]
-    fn build_task<'a>(&self) -> task::Builder<'a, Self> {
-        task::Builder::new(self.clone())
+
+    pub trait Schedule: Sized + Clone {
```
hawkw, line 29: I'm not sure if this is really the right way to seal the trait. I think user code _does_  need to be able to reference the trait name, so that it can write code that is generic over `S: Schedule`, or potentially also use the trait's methods. It just can't be _implemented_ in user code.

In order to ensure that, I think the trait definition of `Schedule` needs to be in the public module, but it needs to require a `Sealed` marker trait that is only visible internally. Like:

```rust
pub trait Schedule: sealed::Sealed + Sized + Clone {
     // ...
}

mod sealed {
    pub trait Sealed {}
}
```

hawkw (changes requested):

jamesmunns:
Closing this, mostly intended to open this as a conversation point, but I'd need to do the sealing differently.

PR #255, opened 2022-07-17
fix(bitfield): respect 32 bit targets

WaffleLapkin:
_Personally, I'd just not allow `usize` at all, but since it already is, we should at least support 16 bit targets..._

I've rearranged primitive `FromBits` impls so that it's [clear that]
- `impl FromBits<T> for {u,i}size` is only supported on platforms with `size_of::<usize>() <= size_of::<T>()`
- `impl FromBits<{u,i}size> for T` is only supported on platforms with `size_of::<T>() <= size_of::<usize>()`

### DIFF
--- a/bitfield/src/lib.rs
+++ b/bitfield/src/lib.rs
@@ -108,30 +108,57 @@ macro_rules! impl_frombits_for_bool {
     }
 }
 
-impl_frombits_for_ty! {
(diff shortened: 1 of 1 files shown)


hawkw (approved):
thanks!

hawkw:
we should maybe remove the usize impls in the next breaking change, i agree that they may not really be worth the effort — who's actually going to use them?

PR #256, opened 2022-07-17
util: add docs and clean up APIs

hawkw:


jamesmunns (approved):
Overall looks great! I love the docs. Made one clippy nit, but otherwise good to merge.

jamesmunns:
(oh, there's also a failing loom test because of const use of Atomic* types)

PR #261, opened 2022-07-23
feat(maitake): implement `JoinHandle`s

hawkw:
This mostly works now, it probably needs more tests

### DIFF
--- a/maitake/src/task.rs
+++ b/maitake/src/task.rs
@@ -14,6 +14,7 @@ pub use self::storage::Storage;
 pub use core::task::{Context, Poll, Waker};
 
 mod builder;
+mod join_handle;
 mod state;
 mod storage;
 
@@ -23,8 +24,9 @@ mod tests;
 use crate::{
     loom::cell::UnsafeCell,
     scheduler::Schedule,
-    task::state::{OrDrop, ScheduleAction, StateCell},
+    task::state::{OrDrop, PollAction, ScheduleAction, StateCell},
     util::non_null,
+    wait::WaitCell,
 };
 
 use core::{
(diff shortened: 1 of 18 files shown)


hawkw:
I think we can move forward with this. There's lots of room for improvement, but it works, and it has _some_ tests...

maitake/src/task.rs
```diff
@@ -161,14 +179,28 @@ pub(crate) struct Header {
...
+    /// The future has completed, and its output is ready to be taken by a
+    /// `JoinHandle`, if one exists.
+    Ready(F::Output),
+    /// The future has completed, and the task's output has been taken or is not
+    /// needed.
+    Joined,
 }
 
-#[derive(Debug)]
+// #[derive(Debug)]
```
jamesmunns, line 192: Remove or restore?

maitake/src/task.rs
```diff
@@ -329,7 +363,7 @@ where
             "Task::poll"
         );
         let mut this = ptr.0.cast::<Self>();
-        test_trace!(task = ?fmt::alt(this.as_ref()));
+        test_debug!(task = ?fmt::alt(this.as_ref()));
```
jamesmunns, line 366: Did you want to keep this elevated?

jamesmunns: (there are a bunch more of these later in the PR too)

jamesmunns (approved):
A couple of nits, but looks super good! Looking forward to using this :D

maitake/src/task.rs
hawkw: oh, i had to add a manual impl of this; i should remove the commented out derive attribute. good catch!

maitake/src/task.rs
hawkw: these used to be logged at `DEBUG` previously, as well, the macro was (confusingly) named `test_trace!` but emitted a `DEBUG` event. the reason a bunch of them changed is because i renamed the macro and added a new `test_trace!` that emits a `TRACE` event.

PR #263, opened 2022-07-25
fix(maitake): fix `WaitCell` returning `Closed` when woken

hawkw:
There's currently a bug in `WaitCell::register_wait` where the `Closed`
error is returned if the `WaitCell` is woken during a register attempt.
I believe this was introduced in PR #223, due to a misunderstanding of
this line:

```rust
// Reset to the WAITING state by clearing everything *except*
// the closed bits (which must remain set).
let state = test_dbg!(self.fetch_and(State::CLOSED, AcqRel));
// The only valid state transition while we were parking is to
// add the CLOSED bit.
debug_assert!(
    state == actual || state == actual | State::CLOSED,
    "state changed unexpectedly while parking!"
);
```

This code does *not* set the `CLOSED` bit if it is not already set, it
just clears all other bits *except* the `CLOSED` bit. So, if the
`CLOSED` bit is not set, it remains unset, but if it is set, it is not
cleared. I think that in #223, this was misunderstood as *setting* the
closed bit, so the `closed()` error was always returned. But, the cell
may not actually have closed.

This branch changes that match arm to return the `closed()` error *if*
the closed bit was actually set by another thread, and `notified()`
otherwise.

### DIFF
--- a/maitake/src/wait/cell.rs
+++ b/maitake/src/wait/cell.rs
@@ -131,6 +131,10 @@ impl WaitCell {
         match test_dbg!(self.compare_exchange(State::PARKING, State::WAITING, AcqRel)) {
             Ok(_) => registered(),
             Err(actual) => {
+                // If the `compare_exchange` fails above, this means that we were notified for one of
+                // two reasons: either the cell was awoken, or the cell was closed.
+                //
+                // Bail out of the parking state, and determine what to report to the caller.
                 test_trace!(state = ?actual, "was notified");
                 let waker = self.waker.with_mut(|waker| unsafe { (*waker).take() });
                 // Reset to the WAITING state by clearing everything *except*
--- a/maitake/src/wait/cell.rs
+++ b/maitake/src/wait/cell.rs
@@ -134,7 +134,9 @@ impl WaitCell {
                 test_trace!(state = ?actual, "was notified");
                 let waker = self.waker.with_mut(|waker| unsafe { (*waker).take() });
                 // Reset to the WAITING state by clearing everything *except*
-                // the closed bits (which must remain set).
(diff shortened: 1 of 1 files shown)


maitake/src/wait/cell.rs
```diff
@@ -134,7 +134,9 @@ impl WaitCell {
                 test_trace!(state = ?actual, "was notified");
                 let waker = self.waker.with_mut(|waker| unsafe { (*waker).take() });
                 // Reset to the WAITING state by clearing everything *except*
-                // the closed bits (which must remain set).
+                // the closed bits (which must remain set). This `fetch_and`
+                // does *not* set the CLOSED bit if it is unset, it just doesn't
+                // clear it.
                 let state = test_dbg!(self.fetch_and(State::CLOSED, AcqRel));
                 // The only valid state transition while we were parking is to
                 // add the CLOSED bit.
```
jamesmunns, line 142: I think this was the actual comment that tripped me up, thinking that the only way that the `self.compare_exchange(State::PARKING, State::WAITING, AcqRel)` could fail would be a closed event.

maitake/src/wait/cell.rs
```diff
@@ -134,7 +134,9 @@ impl WaitCell {
                 test_trace!(state = ?actual, "was notified");
```
jamesmunns, line 134:
```suggestion
                // If the `compare_exchange` fails above, this means that we were notified for one of
                // two reasons: either the cell awoken, or the cell was closed.
                //
                // Bail out of the parking state, and determine what to report to the caller.
                test_trace!(state = ?actual, "was notified");
```

jamesmunns (approved):
Whoops. Good catch @hawkw!

PR #269, opened 2022-07-27
chore: adopt Contributor Covenant as CoC

hawkw:
This branch adds the Contributor Covenant Code of Conduct as Mycelium's
CoC. In practice, this is a very small, personal project without a large
number of contributors, so there hasn't really been any need for a CoC
so far. However, we need one to apply for a Netlify open-source plan and
fix our docs build (which is entirely reasonable!) and Contriibutor
Covenant is widely used so this should be fine.


jamesmunns (approved):
Very +1 from me!

PR #272, opened 2022-07-28
chore(maitake): Fix misc release mode errors

### LINKED ISSUE #273: chore: Add release mode tests
Since parts of mycelium tend to have extensive "debug mode only" features, we should probably also test release mode in CI, otherwise release regressions might slip through.

This caused https://github.com/hawkw/mycelium/pull/272, and at least one other from memory.

jamesmunns:
A couple more issues showed up in release mode only.

Edit:

Added a basic `cargo check --release`, which I verified would have caught this regression.

Closes #273


hawkw (approved):
great, thank you!

PR #276, opened 2022-07-29
perf(maitake): don't use `WaitCell` for join wakers

hawkw:
Currently, `maitake`'s tasks store the `Waker` of any `JoinHandle` that
has polled the task using a `WaitCell`. This is somewhat inefficient, as
`WaitCell`s have their own atomic word of state, but only a couple bits
of state are actually needed to control access to the `Waker`.

Since `Task`s already have their own atomic word of state, we can reduce
the task's heap size significantly by changing from using a `WaitCell`
to just using a couple additional bits of the task's `StateCell` to
track the waiter's state. This gets rid of a word-sized atomic and (when
cache padding is enabled) that atomic's cache-line padding as well. By
storing the waker in a `MaybeUninit` instead of an `Option` (which is
fine here, as we will never set a waker again after consuming it), we
also avoid the `Option`'s enum descriminant...as well as any padding
added by the compiler to align the `WaitCell`.

Adding a test that prints the total size of an empty[^1] `Task`, we see
a significant decrease in the release-mode size of the `Task` struct.

Before:

```shell
:# eliza at noctis in mycelium on eliza/task-sz-test [$]
:; cargo test -p maitake --release --lib -- empty_task_size --show-output
    Finished release [optimized] target(s) in 0.08s
     Running unittests src/lib.rs (target/release/deps/maitake-a93de07a086e60a9)

running 1 test
test task::tests::alloc::empty_task_size ... ok

successes:

---- task::tests::alloc::empty_task_size stdout ----
maitake::task::Task<maitake::task::tests::alloc::NopSchedule, futures_util::future::ready::Ready<()>, maitake::task::storage::BoxStorage>: 384B
futures_util::future::ready::Ready<()>: 1B
task size: 383B
```

After:

```shell
:# eliza at noctis in mycelium on eliza/joinwaker-unsafe-cell [$]
:; cargo test -p maitake --release --lib -- empty_task_size --show-output
    Finished release [optimized] target(s) in 0.08s
     Running unittests src/lib.rs (target/release/deps/maitake-a93de07a086e60a9)

running 1 test
test task::tests::alloc::empty_task_size ... ok

successes:

---- task::tests::alloc::empty_task_size stdout ----
maitake::task::Task<maitake::task::tests::alloc::NopSchedule, futures_util::future::ready::Ready<()>, maitake::task::storage::BoxStorage>: 88B
futures_util::future::ready::Ready<()>: 1B
task size: 87B
```

A reduction of 296 bytes!

Depends on #275.

[^1]: Created by spawning `futures::future::Ready<()>`.

### DIFF
--- a/maitake/src/lib.rs
+++ b/maitake/src/lib.rs
@@ -2,6 +2,8 @@
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg, doc_cfg_hide))]
 #![cfg_attr(docsrs, doc(cfg_hide(docsrs, loom)))]
 #![cfg_attr(not(test), no_std)]
+#![allow(unused_unsafe)]
+
 #[cfg(feature = "alloc")]
 extern crate alloc;
 
--- a/maitake/src/task.rs
+++ b/maitake/src/task.rs
@@ -24,13 +24,7 @@ mod storage;
 #[cfg(test)]
 mod tests;
 
-use crate::{
-    loom::cell::UnsafeCell,
-    scheduler::Schedule,
-    task::state::{OrDrop, PollAction, ScheduleAction, StateCell},
-    util::non_null,
-    wait::WaitCell,
-};
+use crate::{loom::cell::UnsafeCell, scheduler::Schedule, util::non_null};
 
 use core::{
     any::type_name,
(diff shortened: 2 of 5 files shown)


jamesmunns (approved):
Looks great, docs are super readable.

PR #279, opened 2022-07-30
perf(maitake): fix excessive waker vtable monomorphization

### LINKED ISSUE #274: maitake: We could monomorphize task vtables less
From a hardware no_std build of mnemos using maitake, I had the following cargo bloat output:

https://gist.github.com/jamesmunns/b4feb5c52d144ec683523f33598aa49f

@hawkw noted we could monomorphize smarter:

> maitake is monomorphizing the waker vtable methods
>
> per task
>
> but the waker vtable only needs to be monomorphized per scheduler
>
> it doesn't care about the task's future/storage, only the scheduler
>
> (it cares about the future bc the tracing event in there records the future type, but we could just remove that)

hawkw:
Currently, the `RawWakerVTable` methods in `maitake` are implemented as
associated functions on the `Task` struct. Because `Task` is generic
over the task's `Future`, `Schedule`, and `Storage`, that means these
functions are monomorphized separately for *each* `Future` type that's
spawned by a program, resulting in increased binary size. Because the
waker vtable methods don't actually interact with the task's `Future` or
touch its allocation, however, they don't actually *need* to be
monomorphic over the `F` or `STO` type parameters for the `Future` and
`Storage` types. The only type parameter they actually care about is the
`S`-typed `Schedule` instance. We can decrease the size of binaries
using `maitake` by making sure these functions are monomorphized only
over the `S`-typed scheduler, and not the `Future` or `Storage` types.

This branch fixes the excessive monomorphization of `RawWakerVTable`
methods by factoring out a separate `Schedulable` struct which contains
a task's `Header` and `S`-typed `Schedule` reference. The `Schedulable`
struct can be used as a target to cast a type-erased task pointer to
when only the `Schedule` type is needed. The waker vtable methods are
moved to the `Schedulable` struct, and should therefore be generated
only once per `Schedule` type in a program.

The `Schedulable` struct is `#[repr(C)]` and should be laid out
correctly so that casting a `Header` pointer to a `Task<S, F, STO>`
pointer is still valid. I've updated the test that task layouts are
reasonable to also assert this.

As per this gist from @jamesmunns, which shows `cargo bloat` output
before and after this change, there is now only a single
monomorphization of each waker vtable function:
https://gist.github.com/jamesmunns/b4feb5c52d144ec683523f33598aa49f

Fixes #274


hawkw:
This is a draft PR as I want to write some more docs for it, but the code change is reviewable now.

jamesmunns (approved):
Doc comments look good, and I've tested approximately this in the simulator at least.

PR #282, opened 2022-07-30
feat(maitake): add `Schedule::current_task`

### LINKED ISSUE #277: maitake: `Schedule::current_task` method
It would be nice to have a function for getting the current task. This can be used by leaf futures if they want to self-wake, access information about what task they're being polled from, such as for debugging (see #265 and #267) , or forcibly cancel the current task (see #264).

`maitake` itself can't implement an "ambiently available" `task::current()` function, because it doesn't know about "locality"[^1]. But, we can add a method to the `Scheduler` types for accessing the task that scheduler is currently polling. Then, the OS can use this to implement a `task::current()` by finding the current core's scheduler (or picking the global one if single-core).

I imagine this would have a signature like
```rust
pub fn current_task(&self) -> Option<TaskRef>
```

We would probably have to store the task ptr in an `AtomicPointer` in the scheduler, and load that and that into a `TaskRef` if someone wants it.

[^1]: Are we single-core, and there's a single global scheduler? Or are we multi-core, and each core has its own scheduler? And if so, how do I access the scheduler for the current core? Those are all OS-level concerns that `maitake` can't handle.

hawkw:
This adds an initial implementation of a `Schedule::current_task` method
for cloning the currently-polled task. This implementation is somewhat
inefficient, as it has to perform an additional ref count clone/drop
cycle on every poll. This is unfortunately necessary because the current
task variable is not unset until the `TaskRef::poll` method returns, and
that method can drop the task, so we have to keep the task alive as long
as it's accessible through the scheduler (or else we risk handing out
`TaskRef`s pointed at deallocated tasks).

We can probably improve this by removing the unnecessary ref clone/drop
cycle, if we refactored how polling works a bit. If the scheduler unset
the current task variable _before_ a task is dropped, we wouldn't need
to clone the task before storing it in the current task variable...but,
this implementation works for now.

Closes #277

### DIFF
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -1,5 +1,8 @@
-use crate::task::{self, Header, JoinHandle, Storage, TaskRef};
-use core::{future::Future, pin::Pin};
+use crate::{
+    loom::sync::atomic::{AtomicPtr, Ordering},
+    task::{self, Header, JoinHandle, Storage, TaskRef},
+};
+use core::{future::Future, pin::Pin, ptr};
 
 use cordyceps::mpsc_queue::MpscQueue;
 
@@ -10,6 +13,7 @@ pub struct StaticScheduler(Core);
 #[derive(Debug)]
 struct Core {
     run_queue: MpscQueue<Header>,
+    current_task: AtomicPtr<Header>,
     // woken: AtomicBool,
 }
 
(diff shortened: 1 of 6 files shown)


hawkw:
i think that if the scheduler core's `current_task` atomic got unset here instead https://github.com/hawkw/mycelium/blob/d02a34310d4129310d8c883b3f0a1ccd381558af/maitake/src/task.rs#L498 (between the actual poll and handling the post-poll action), it wouldn't need to *own* a ref clone. so, that would just mean re-structuring the poll code a bit so that we can touch the scheduler there.

PR #285, opened 2022-07-31
feat(maitake): track additional scheduler metrics

hawkw:
This branch adds additional scheduler metric counters, including
counters of woken tasks, spawned tasks, etc.

We may want to feature-flag scheduler metrics, since they may have a
(very minor) performance cost. Currently, the `completed` counter is
used in tests, but we could feature-flag the rest. @jamesmunns what do
you think? Should these counters always be enabled?

### DIFF
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -1,5 +1,5 @@
 use crate::{
-    loom::sync::atomic::{AtomicPtr, Ordering},
+    loom::sync::atomic::{AtomicPtr, AtomicUsize, Ordering::*},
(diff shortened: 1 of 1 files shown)


hawkw:
@jamesmunns says on discord that he's +1 on having these always-on for now, we can figure out feature flagging (if necessary) before publishing.

jamesmunns (approved):
Super awesome. I'll add these as something you can view on my shell interface 💯

hawkw:
> Super awesome. I'll add these as something you can view on my shell interface 100

awesome, they also get logged at `DEBUG` at the end of a tick.

PR #288, opened 2022-08-02
chore: add a `justfile` for automating test workflows

hawkw:
This commit adds a `justfile` for the [Just] command runner. This
performs most of what was previously automated in a big pile of shell
scripts --- hopefully it's a bit nicer to have all this stuff in one
place.

A potential downside of this change is that it does introduce an
additional dependency on `just` for users who want to use the Just
recipes. We could consider alternative approaches like trying to
automate all of these workflows with `cargo` subcommands or something...

[Just]: https://just.systems


hawkw:
@jamesmunns wdyt?

jamesmunns:
I'm generally pro-`just`, and it's not an unreasonable dep IMO.

I think the other "reasonable" approach would be to use the `xtask` approach, which I was thinking about setting up for mnemos (or using `just` if that went sub-par, like you did here).

No strong feelings here from me :)

hawkw:
> I think the other "reasonable" approach would be to use the `xtask` approach, which I was thinking about setting up for mnemos (or using `just` if that went sub-par, like you did here).

Yeah...we already have a custom cargo subcommand in-repo, with `inoculate`, which is actually needed for compiling the bootloader and stuff. So I also considered just hanging more stuff off of that.

The main advantages of `just` here is that a lot of these recipes are just very simple "run this test command, and also this test command, and also this test command", which is easily expressed in a command-runner tool like `just` but a bit more boilerplate to write in Rust, and that it's a lot easier to port the existing Bash scripts which can just be pasted into the `justfile` (rather than having to RIIR them). On the other hand, I kind of like Cargo subcommands/`xtask`s because it's nice to not require any additional dependencies for someone to be able to check out and build the project, and also because it's maybe a _bit_ of a pain to have two different things you type to run builds. 

But, in this case, the `justfile` is only here to bundle together a bunch of different Cargo commands, basically --- `just` isn't a necessary dep for building the Mycelium kernel, which can be done entirely from Cargo; it's just providing additional convenience for testing workflows...

jamesmunns (approved):
If this works today, go for it! Lots of other options if it ends up being a pain :)

Issue #199, opened 2022-06-05
util: parameterize mutex blocking behavior


hawkw:
this was originally suggested by @jamesmunns in a comment on #191:

> Misc note (can totally be a future PR): It might be nice to be able to modify the lock behavior somehow.
>
> For me, any lock contention is likely to be a programming bug, so `.try_lock().unwrap()` is likely to make a "louder"/easier to catch case for me.
>
> __Originally posted by @jamesmunns in https://github.com/hawkw/mycelium/pull/191#issuecomment-1146854154__

i think the nicest way to do this is add some kind of trait to `mycelium_util` that represents how a lock would block, and make it generic over a type parameter of that trait. by default, it can spin, but with the appropriate marker type, a lock could be configured to panic any time it would block.

this could also be useful for passing in a platform-specific way to disable interrupts while a lock is held, as discussed in #88.

hawkw:
I'm not actually sure if doing this using a trait is the best approach. You would ideally want to be able to disable spinlocks in dependencies as well as your own code (e.g. in `maitake`), and if this was done using a trait, libraries using mutexen would need to parameterize basically everything in the library over that trait. In some cases, such as when the mutex is in a `static` internal to the library, this can't be done that easily.

Perhaps a better approach would be to introduce a `RUSTFLAGS` cfg like `mycelium_util_single_core` or something that internally changes the mutex behavior. It would have to be set via `RUSTFLAGS` rather than via a feature flag, since we would only want the top level project (the OS/app) to set it, and not have library dependencies set it. 

Unfortunately, this approach would probably not work for #88, but it would give us a way to disable spinlocks on single-core-only targets.

@jamesmunns what do you think?

hawkw:
oh hey this is what #482 was all about!

PR #297, opened 2022-08-11
feat(maitake): add `TaskId`s

hawkw:


maitake/src/scheduler.rs
```diff
@@ -204,7 +204,7 @@ impl Core {
         };
 
         for task in self.run_queue.consume() {
-            let _span = debug_span!("poll", ?task).entered();
+            let _span = debug_span!("poll", ?task, task.tid = %task.id()).entered();
```
jamesmunns, line 207: Would it make sense to have an "as integer" method here? Otherwise we'll need to do `atoi` every time to format the task ID into a string.

maitake/src/task.rs
```diff
@@ -489,10 +497,22 @@ where
         }
     }
 
+    /// Returns a [`TaskId`] that uniquely identifies this [task].
```
jamesmunns, line 500: Doc failure here

maitake/src/task/id.rs
```diff
@@ -0,0 +1,82 @@
+use core::fmt;
+
+/// A unique identifier for a running [task].
+///
+/// A `TaskId` is an opaque object that uniquely identifies each [task]
+/// created during the lifetime of a program. `TaskIds`s are guaranteed not to
```
jamesmunns, line 6: Might wait to slightly weaken this claim? Very long-running hardware could potentially wrap a u32.

maitake/src/task.rs
jamesmunns: (also one later, also referencing `task`)

maitake/src/task/id.rs
hawkw: This is the guarantee `std` makes for `ThreadId`s, and it does this by using a mutexed `u64` on platforms without 64-bit atomics. We could either weaken the guarantee, or use a similar approach to ensure that task IDs are always `u64`s. We currently assert that the task ID counter doesn't wrap, so if we think there's the potential of wrapping on 32-bit platforms, we probably should also remove that assertion: https://github.com/hawkw/mycelium/blob/ec15a80e99601a52e826536f0b6773ccd32a89ac/maitake/src/task/id.rs#L32

This is the real reason I liked using addrs for task IDs, it has weaker uniqueness guarantees because an address might be reused, but it _does_ guarantee that task IDs will never collide with the IDs of *other running tasks*, which is a concern if counters can wrap . Tasks with low IDs are spawned early in the program's lifetime, and might be more likely to be long running tasks, so if (say) task `1` is a long-lived task in a program that runs for a long time and spawns a bunch of short lived tasks constantly, if a 32-bit counter wraps, we might see ID `1` assigned to a second task _while_ the original task `1` is still running. I would personally prefer IDs that are always guaranteed to be unique regardless of word size, even if they're _not_ guaranteed to not be reused... :woman_shrugging:

hawkw:
@jamesmunns i've updated this branch to always use 64-bit IDs, using a spinlock on platforms without 64-bit atomics.

jamesmunns:
Sounds good, I'm going to mention #199, since this might be another place for it. Spinlock works for me, like we discussed, using scheduler APIs in unidirectionally preemptive situations (like in interrupts) is already A Bad Idea for other reasons, so I don't see any reason to block on that.

jamesmunns (approved):

PR #301, opened 2022-08-29
feat(maitake): add `wait::Semaphore`

### LINKED ISSUE #299: maitake: add an async semaphore
It would be nice to have an async semaphore implementation for `maitake`.

A semaphore is a useful synchronization type for things like rate-limiting async tasks. It could also be used as a lower-level primitive to implement things like read-write locks[^1] and channels[^2].

We can probably do something very similar to the implementation I wrote for Tokio in tokio-rs/tokio#2325, although it should be possible to simplify some code that was only necessary to maintain Tokio's API surface.

[^1]: a rwlock can be modeled by a semaphore with _n_ permits (where _n_ is the maximum number of concurrent readers); each reader must acquire a single permit, while a writer must acquire _n_ permits).

[^2]: a bounded MPSC channel of capacity _n_ can be implemented using a semaphore with _n_ permits, where each producer must acquire a single permit to write, and every time a message is consumed, the reader releases a permit to the writers.

hawkw:
A semaphore is a useful synchronization type for things like
rate-limiting async tasks. It could also be used as a lower-level
primitive to implement things like read-write locks[^1] and
channels[^2].

This branch adds an asynchronous semaphore implementation in
`maitake::wait`. The implementation is based on the implementation I
wrote for Tokio in tokio-rs/tokio#2325, with some code simplified a bit
as it was not necessary to maintain Tokio's required API surface.

Closes #299 

[^1]: a rwlock can be modeled by a semaphore with _n_ permits (where _n_
    is the maximum number of concurrent readers); each reader must
    acquire a single permit, while a writer must acquire _n_ permits).

[^2]: a bounded MPSC channel of capacity _n_ can be implemented using a
    semaphore with _n_ permits, where each producer must acquire a
    single permit to write, and every time a message is consumed, the
    reader releases a permit to the writers.


hawkw:
This is a draft primarily because I'd like to add docs.

hawkw:
@jamesmunns any opinions as to whether this belongs in `maitake::wait` or `maitake::sync`? Or, should we just combine those modules?

PR #306, opened 2022-08-31
chore(maitake): waitcell cleanup

hawkw:
This is a separate branch from #304 because it actually changes API naming.

cc @jamesmunns any objections to this before i merge it? it will break some API type naming.


jamesmunns (approved):
No objections from me!

PR #311, opened 2022-09-03
feat(maitake): add `time` module with timer wheel and futures

hawkw:
This branch introduces a new `maitake::time` module in `maitake`. This
module contains `Sleep` and `Timeout` futures, and a `Timer` type which
is necessary to drive these futures.

The `Timer` type is an implementation of a [hierarchical timer
wheel][1], which can track and notify a large number of `Sleep` futures
efficiently. Because `maitake` is platform-agnostic, the timer wheel is
designed to be driven by an external time source, such as a timer
interrupt, or some way of reading a current timestamp. The base
granularity of the timer wheel is configured when the wheel is
constructed.

Timer wheels can be constructed in `const fn`s and require no allocation
to use. Therefore, I've also added support for setting an
ambiently-available global timer.

There are a lot of comments in the code, so I'm not going to go into
*too* much detail in this commit message. See the [`time` module's
docs][2] for details.

In a subsequent PR, I also intend to add support for using the timer as
a "clock", with an `Instant` type and a `now` function. But, I figured
that was best left for a follow-up PR...especially since I'm working on
something *really* cursed in order to have an efficient 64-bit timestamp
`now` on platforms that don't have 64-bit atomics.

[1]:
    http://www.cs.columbia.edu/~nahum/w6998/papers/sosp87-timing-wheels.pdf
[2]:
    https://deploy-preview-311--mycelium-docs.netlify.app/maitake/time/index.html


hawkw:
@jamesmunns okay, this branch is *finally* ready, if you're interested in giving it a review (no pressure, of course)!

PR #334, opened 2022-10-08
[wip] feat(kernel): use a sponge to generate nicer RNGs

hawkw:
Per a suggestion from @hdevalence. This isn't done yet because...it turns out our current QEMU configuration doesn't expose RDRAND support to the guest, and I don't know if timestamp-based entropy added to the sponge at a fixed interval is a good idea...


hawkw:
It would be nice to get this merged before implementing a `/dev/random`-like API for userspace, so that we can generate CSPRNGs nicely using the sponge...

hawkw:
Suggestions from @hdevalence:
> like there’s some kind of contortions to make like a fake RNG in order to use the merlin api or whatever
instead of doing that, you could just vendor the strobe implementation from the merlin crate
the code isn’t going to change
and then you can use the strobe state directly, and you can absorb the entropy in a way that provides forward secrecy
basically, when you’re ingesting new entropy, you want to be in K mode
so it overwrites rather than xoring

hawkw:
**from henry:

> OK so here’s what i’d do
> make a 
> ```rust
> struct MyceliumRng {
>     inner: Strobe128
> }
> ```
> then have
> ```rust
> MyceliumRng::new(seed: &[u8])
> MyceliumRng::reseed(&mut self, data: &[u8])
> MyceliumRng::fork(&mut self) -> MyceliumRng
>
> impl RngCore for MyceliumRng
> ```
> - `new` should initialize the struct and then immediately `reseed` it with the data
> - `reseed` should do a strobe KEY operation with the provided data
> - the `RngCore` impl should use the PRF operation to fill bytes
> the `fork` operation is a little tricky. you want to make sure that the two forks have different transcripts
> because it has to be different regardless of the state of the world
> i think you want to do something like, clone the existing state, then reseed both the existing and the cloned state with different data
> eg `[0]` and `[1]`
> oh, okay
> this way, if i do `.fork()` a bunch of times, i can never accidentally get the same outputs, because the state is bound to a transcript of that instances path through the fork tree
> remember it doesn’t need extra entropy, it just has to have different transcript data
>
> the idea is that `fork` is like `Clone` but safe
> ideally it would actually Be the `Clone` impl
> but that’s not possible because you need to mutate *both* instances
> and the reason you need to do that is so if you fork twice you don’t get two copies of “forked from X”

PR #439, opened 2023-06-01
feat(maitake): add local (`!Send`) scheduler types

hawkw:
Currently, `maitake`'s scheduler types require tasks to be `Send`. This is because the scheduler itself is `Send` and `Sync`, and it may be ticked from more than one thread/CPU core. Ticking a scheduler which contains `!Send` tasks spawned from one thread on another thread is essentially equivalent to transferring ownership of those `!Send` tasks across threads, so a `Send + Sync` scheduler cannot execute `!Send` futures.

This branch introduces new `LocalScheduler` and `LocalStaticScheduler` types for executing `!Send` futures. Other than the fact that they can spawn futures which are not `Send`, and are themselves not `Send` or `Sync`, they are functionally the same as the existing `Scheduler` and `StaticScheduler` types. Additionally, in order to support other threads to spawn tasks which *are* `Send` on a local scheduler instance, new `LocalSpawner` and `LocalStaticSpawner` types can be created and shared across thread boundaries. These spawner types can *only* spawn `Send` tasks on the scheduler, and are not capable of ticking the scheduler. Therefore, they cannot accidentally access a `!Send` future.

### DIFF
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -1648,6 +1648,7 @@ feature! {
         /// [run-loops]: crate::scheduler#executing-tasks
         #[inline]
         #[track_caller]
+        #[cfg(not(loom))]
         pub fn spawn<F>(&self, future: F) -> JoinHandle<F::Output>
         where
             F: Future + Send +'static,
@@ -1686,6 +1687,7 @@ feature! {
         /// [run-loops]: crate::scheduler#executing-tasks
         #[inline]
         #[track_caller]
+        #[cfg(not(loom))]
         pub fn spawn_allocated<F>(&self, task: Box<Task<LocalScheduler, F, BoxStorage>>) -> JoinHandle<F::Output>
         where
             F: Future + Send + 'static,
--- a/maitake/src/scheduler.rs
+++ b/maitake/src/scheduler.rs
@@ -679,7 +679,7 @@ impl LocalStaticScheduler {
 
     /// Spawn a pre-allocated, ![`Send`] task.
     ///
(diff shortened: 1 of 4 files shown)


hawkw:
In theory, it would be possible to change the internals of a `!Send` scheduler to reduce the use of atomics for task reference counts and task states, as well. This would potentially improve performance over the implementation that uses atomics, at least a little bit. However, this branch does *not* do that, because it would basically require us to essentially maintain two separate versions of the entire `task` module. We could address this later if it's important to us.

jamesmunns (approved):
Very much out of the loop of the internals of maitake, but the docs are very comprehensive, and I feel like I have a decent idea of how things should be used!

I wonder if there is SOME way to abstract over the Send/!Send, like some kinda side marker trait the Scheduler could be generic over, with some fancy associated type bounds, so that it ends up being something like `pub struct LocalScheduler(Scheduler<Local>);` and `pub struct FancyScheduler(Schedule<Fancy>)`. But that might be too astronaut architect'y.

Excited to try this out in mnemos!

hawkw:
> I wonder if there is SOME way to abstract over the Send/!Send, like some kinda side marker trait the Scheduler could be generic over, with some fancy associated type bounds, so that it ends up being something like `pub struct LocalScheduler(Scheduler<Local>);` and `pub struct FancyScheduler(Schedule<Fancy>)`. But that might be too astronaut architect'y.

Yeah, I also wondered about this. In practice, the way we reuse the common code is just that all the schedulers own a `Core`, which has the actual scheduling logic...it would be nice to not have to write out the same method signatures so many times, but :woman-shrugging:

PR #441, opened 2023-06-04
feat(maitake): return next deadline from `Timer::advance`

### LINKED ISSUE #440: maitake::time: public API for getting the time until the next deadline in the wheel
currently, the timer wheel can be advanced to a new "now" time, firing any timers whose deadline is <= now. there isn't currently an API that just gives you the next deadline, even if it's after the current now.

it would be nice to add this, so that we can have a way to sleep until the next deadline using a hardware timer. this way, when there's no work to do, we can put the system in a low power state until the next timer fires, get woken by an IRQ, and then advance the wheel.

hawkw:
On hardware platforms with a variable-duration one-shot timer interrupt, the timer interrupt can be configured to fire at the next deadline on the timer wheel, rather than at a fixed interval. This is more efficient than advancing the wheel every time a periodic timer fires, as in many cases, most of those interrupts may not correspond to the deadline of any pending timer. It also permits more precise timing, as a periodic timer interval may "overshoot" the deadline of a particular sleep future. Currently, the `maitake::time::Timer` API doesn't expose the deadline of the next firing timer, though, so it's not currently possible to implement this pattern.

This branch changes the `Timer::advance`, `Timer::advance_ticks`, `Timer::force_advance`, and `Timer::force_advance_ticks` methods to return a `Turn` structure, which described what occurred while advancing the timer. This is very similar to the `Tick` structure returned by `Scheduler::tick`. Among other data, the `Turn` struct contains the time until the next expiring deadline. This can be used to configure the hardware timer in oneshot mode.

We could probably also add a method on `Timer` to just return the next deadline without advancing the wheel. However, this would currently require acquiring the lock. Since advancing the timer already requires locking the wheel, and we will generally care about the next deadline when we have just advanced the wheel, it seems more efficient to return it from `advance` (and friends), without re-locking.

Closes #440

### DIFF
--- a/maitake/src/time/timer.rs
+++ b/maitake/src/time/timer.rs
@@ -100,6 +100,33 @@ use self::sleep::Sleep;
 /// form of runtime bookkeeping action. For example, the timer can be advanced
 /// in a system's run loop every time the [`Scheduler::tick`] method completes.
 ///
+/// ### Periodic and One-Shot Timer Interrupts
+///
+/// Generally, hardware timer interrupts operate in one of two modes: _periodic_
+/// timers, which fire on a regular interval, and _one-shot_ timers, where the
+/// timer counts down to a particular time, fires the interrupt, and then stops
+/// until it is reset by software. Depending on the particular hardware
+/// platform, one or both of these timer modes may be available.
+///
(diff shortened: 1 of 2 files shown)


jamesmunns (approved):
This is very cool, thank you :D

PR #444, opened 2023-06-18
Actually hold on to the test subscriber

jamesmunns:


jamesmunns:
@hawkw I dunno if this was intentional for some other reason, but it *seems* like the "init tracing" helper returned a guard that the caller is supposed to hold, but none (or not all?) of the tests that called it actually held it.

This makes a single return type (because rustc was Big Mad when I used `Box<dyn Drop>`) based on the features, and marks it as must use.

If you want me to rebase this on `main` instead of your test branch, lemme know.

hawkw (approved):
lol, my bad — thanks for fixing this. i would prefer to merge the PR into `main`, and rebase my test PR onto it, if you don't mind?

jamesmunns:
No worries! Replaced by #445.

PR #456, opened 2023-07-24
feat(maitake): add `WaitCell::subscribe` and `poll_wait`

hawkw:
This branch rewrites the `maitake::sync::WaitCell` type a bit to fix
potential lost wakeups accidentally introduced in #453. In addition, it
changes the API a bit:
- Replace `register_wait` with `poll_wait`, which I think is a bit
  nicer,
- Removed the `RegisterError::Waking` variant since `poll_wait` returns
  `Ready` if the cell was woken, and renamed `RegisterError` to just
  `Error`,
- Introduced a `WaitCell::subscribe` method to eagerly pre-register the
  task's `Waker` in the `WaitCell`. This is intended to be used to
  register the task to be woken *before* performing an operation that
  results in a wakeup, such as enabling an interrupt that will wake the
  task.

Closes #455

### DIFF
--- a/maitake/src/sync/wait_cell.rs
+++ b/maitake/src/sync/wait_cell.rs
@@ -627,6 +627,39 @@ mod tests {
     }
 
     #[test]
+    fn wake_debounce() {
+        let _trace = crate::util::test::trace_init();
+        let cell = Arc::new(WaitCell::new());
+
+        let mut task = task::spawn({
+            let cell = cell.clone();
+            async move {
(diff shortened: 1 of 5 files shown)


jamesmunns (approved):
Looks good to me, one docs nit, and there are also docs failures.

hawkw:
okay, you know what? this PR was the worst thing that's ever happened to me. i'm going to give up on trying to write a test to reproduce the last problem.

PR #478, opened 2024-07-06
Address lint ambiguous_wide_pointer_comparisons

jamesmunns:
Noticed this warning:

```
warning: ambiguous wide pointer comparison, the comparison includes metadata which may not be expected
   --> /Users/james/personal/mycelium/cordyceps/src/list.rs:497:12
    |
497 |         if head == tail {
    |            ^^^^^^^^^^^^
    |
    = note: `#[warn(ambiguous_wide_pointer_comparisons)]` on by default
help: use `std::ptr::addr_eq` or untyped pointers to only compare their addresses
    |
497 |         if std::ptr::addr_eq(head.as_ptr(), tail.as_ptr()) {
    |            ++++++++++++++++++    ~~~~~~~~~~     ++++++++++
help: use explicit `std::ptr::eq` method to compare metadata and addresses
    |
497 |         if std::ptr::eq(head.as_ptr(), tail.as_ptr()) {
    |            +++++++++++++    ~~~~~~~~~~     ++++++++++

warning: ambiguous wide pointer comparison, the comparison includes metadata which may not be expected
   --> /Users/james/personal/mycelium/cordyceps/src/list.rs:966:13
    |
966 |             splice_start == splice_end
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
help: use `std::ptr::addr_eq` or untyped pointers to only compare their addresses
    |
966 |             std::ptr::addr_eq(splice_start.as_ptr(), splice_end.as_ptr())
    |             ++++++++++++++++++            ~~~~~~~~~~           ++++++++++
help: use explicit `std::ptr::eq` method to compare metadata and addresses
    |
966 |             std::ptr::eq(splice_start.as_ptr(), splice_end.as_ptr())
    |             +++++++++++++            ~~~~~~~~~~           ++++++++++
    ```

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -494,7 +494,7 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         let tail_links = unsafe { T::links(tail) };
         let head_links = unsafe { head_links.as_ref() };
         let tail_links = unsafe { tail_links.as_ref() };
-        if head == tail {
+        if core::ptr::addr_eq(head.as_ptr(), tail.as_ptr()) {
             assert_eq!(
                 head_links, tail_links,
                 "{name}if the head and tail nodes are the same, their links must be the same"
@@ -963,7 +963,7 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         let start_links = T::links(splice_start).as_mut();
         let end_links = T::links(splice_end).as_mut();
         debug_assert!(
-            splice_start == splice_end
+            core::ptr::addr_eq(splice_start.as_ptr(), splice_end.as_ptr())
                 || (start_links.next().is_some() && end_links.prev().is_some()),
             "splice_start must be ahead of splice_end!\n   \
             splice_start: {splice_start:?}\n    \
--- a/maitake-sync/Cargo.toml
+++ b/maitake-sync/Cargo.toml
@@ -17,7 +17,7 @@ categories = [
     "concurrency"
 ]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [features]
--- a/maitake/Cargo.toml
+++ b/maitake/Cargo.toml
@@ -17,7 +17,7 @@ categories = [
     "async",
 ]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
 
 # NOTE! As a workaround for https://github.com/rust-lang/rust/issues/97708,
 # maitake uses some `no_mangle` functions. This will cause linker errors if
--- a/util/Cargo.toml
+++ b/util/Cargo.toml
@@ -3,7 +3,7 @@ name = "mycelium-util"
 version = "0.1.0"
 authors = ["Eliza Weisman <eliza@elizas.website>"]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
 readme = "README.md"
 
 # See more keys and their definitions at
--- a/cordyceps/Cargo.toml
+++ b/cordyceps/Cargo.toml
@@ -11,7 +11,7 @@ homepage = "https://mycelium.elizas.website"
 readme = "README.md"
 keywords = ["intrusive", "no_std", "list", "queue", "lock-free"]
 categories = ["data-structures", "no-std"]
-rust-version = "1.61.0"
+rust-version = "1.76.0"
 
 [features]
 default = []

### CODE REVIEW

**Review by hawkw — COMMENTED** [review]
This is a MSRV bump, since `core::ptr::addr_eq` was added in Rust 1.76.0, and `cordyceps` currently [claims a minimum Rust version of 1.61.0](https://github.com/hawkw/mycelium/blob/e456145e84ad6339752251318e6cf5d296b65d51/cordyceps/Cargo.toml#L14).

I'm not super fussed about taking the MSRV bump, but if we do, it would be good to make sure to change the Cargo.toml `rust-version` field as part of this commit, as well.

Alternatively, we could perform the same address-without-metadata comparison by writing 
```rust
head.as_ptr() as *const () == tail.as_ptr() as *const ()
```
and
```rust
splice_start.as_ptr() as *const () == splice_end.as_ptr() as *const ()
```
which is basically what `ptr::addr_eq` does internally, but is permitted on older Rust versions. If we go that route, we can avoid the MSRV bump, but it might be worth adding a comment that the `as *const ()` casts are to ignore the wide pointer metadata, so that a future reader doesn't change them back...

hawkw:
Ugh, it looks like some of the maitake tests are not feature flagged correctly: https://github.com/hawkw/mycelium/actions/runs/9821097757/job/27116399920?pr=478 

I can fix that separately...

jamesmunns:
Updated MSRV of `cordyceps`, and all crates that I could see that depend on cordyceps.

maitake-sync/Cargo.toml
```diff
@@ -17,7 +17,7 @@ categories = [
     "concurrency"
 ]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
```
hawkw, line 20: I don't actually think we need to do the MSRV bump for `maitake-sync`. Since the `cordyceps` changes will be released in a semver-compatible version, and `maitake-sync`'s minimum `cordyceps` version isn't changing, users of `maitake-sync` can continue using the current cordyceps version rather than picking up the one with the new MSRV --- either by checking in a lockfile, or by explicitly pinning the `cordyceps` version in their Cargo.toml. 

IMHO, we would only need to change `maitake-sync`'s MSRV if we were changing the dependency spec for `cordyceps` to not permit versions that compile on 1.61.0.

cordyceps/src/list.rs
```diff
@@ -494,7 +494,7 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         let tail_links = unsafe { T::links(tail) };
         let head_links = unsafe { head_links.as_ref() };
         let tail_links = unsafe { tail_links.as_ref() };
-        if head == tail {
+        if core::ptr::addr_eq(head.as_ptr(), tail.as_ptr()) {
```
hawkw, line 497: nit, take it or leave it: `core::ptr` is already imported in this file: https://github.com/hawkw/mycelium/blob/ed9e49b520b574298ae777a06bd8b5fc125c05ac/cordyceps/src/list.rs#L13

so we could drop the `core` prefix:
```suggestion
        if ptr::addr_eq(head.as_ptr(), tail.as_ptr()) {
```

cordyceps/src/list.rs
```diff
@@ -963,7 +963,7 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         let start_links = T::links(splice_start).as_mut();
         let end_links = T::links(splice_end).as_mut();
         debug_assert!(
-            splice_start == splice_end
+            core::ptr::addr_eq(splice_start.as_ptr(), splice_end.as_ptr())
```
hawkw, line 966: similarly, we could write:
```suggestion
            ptr::addr_eq(splice_start.as_ptr(), splice_end.as_ptr())
```

maitake/Cargo.toml
```diff
@@ -17,7 +17,7 @@ categories = [
     "async",
 ]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
```
hawkw, line 20: again, i think the MSRV bump isn't strictly necessary unless we are also changing the cordyceps dep

util/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "mycelium-util"
 version = "0.1.0"
 authors = ["Eliza Weisman <eliza@elizas.website>"]
 edition = "2021"
-rust-version = "1.61.0"
+rust-version = "1.76.0"
```
hawkw, line 6: and again, i don't think this is *really* necessary...although it's also fine since this isn't a published crate.

hawkw (changes requested):
Some smallish nits. I'd like to merge this, though!

PR #479, opened 2024-07-06
Add `wait_for` and `wait_for_value` to WaitCell and WaitQueue

jamesmunns:
Implementing the interface that @hawkw so wisely suggested

### DIFF
--- a/maitake-sync/src/wait_cell.rs
+++ b/maitake-sync/src/wait_cell.rs
@@ -335,7 +335,7 @@ impl WaitCell {
     ///
     /// Consider using [`Self::wait_for_value()`] if your function does return a value.
     ///
-    /// * Returns `Ok(T)` if the closure returns `Some(T)`.
+    /// * Returns `Ok(())` if the closure returns `true`.
     /// * Returns `Err(Closed)` if the [`WaitCell`] is closed.
     pub async fn wait_for<F: FnMut() -> bool>(&self, mut f: F) -> Result<(), Closed> {
         loop {
--- a/maitake-sync/src/wait_cell.rs
+++ b/maitake-sync/src/wait_cell.rs
@@ -327,6 +327,46 @@ impl WaitCell {
         }
     }
 
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for_value()`] if your function does return a value.
+    ///
+    /// * Returns `Ok(T)` if the closure returns `Some(T)`.
+    /// * Returns `Err(Closed)` if the [`WaitCell`] is closed.
+    pub async fn wait_for<F: FnMut() -> bool>(&self, mut f: F) -> Result<(), Closed> {
+        loop {
+            let wait = self.subscribe().await;
+            if f() {
+                return Ok(());
+            }
+            wait.await?;
+        }
+    }
+
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for()`] if your function does not return a value.
+    ///
+    /// * Returns `Ok(T)` if the closure returns `Some(T)`.
+    /// * Returns `Err(Closed)` if the [`WaitCell`] is closed.
+    pub async fn wait_for_value<T, F: FnMut() -> Option<T>>(&self, mut f: F) -> Result<T, Closed> {
+        loop {
+            let wait = self.subscribe().await;
+            if let Some(t) = f() {
+                return Ok(t);
+            }
+            wait.await?;
+        }
+    }
+
     // TODO(eliza): is this an API we want to have?
     /*
     /// Returns `true` if this `WaitCell` is [closed](Self::close).
--- a/maitake-sync/src/wait_cell.rs
+++ b/maitake-sync/src/wait_cell.rs
@@ -345,7 +345,7 @@ impl WaitCell {
     /// In particular, this can be used in cases where interrupts or events are used
     /// to signify readiness or completion of some task, such as the completion of a
     /// DMA transfer, or reception of an ethernet frame. In cases like this, the interrupt
-    /// can wake the queue, allowing the polling function to check status fields for
+    /// can wake the cell, allowing the polling function to check status fields for
     /// partial progress or completion.
     ///
     /// Consider using [`Self::wait_for_value()`] if your function does return a value.
@@ -419,7 +419,7 @@ impl WaitCell {
     /// In particular, this can be used in cases where interrupts or events are used
     /// to signify readiness or completion of some task, such as the completion of a
     /// DMA transfer, or reception of an ethernet frame. In cases like this, the interrupt
-    /// can wake the queue, allowing the polling function to check status fields for
+    /// can wake the cell, allowing the polling function to check status fields for
     /// partial progress or completion, and also return the status flags at the same time.
     ///
     /// Consider using [`Self::wait_for()`] if your function does not return a value.
--- a/maitake-sync/src/wait_cell.rs
+++ b/maitake-sync/src/wait_cell.rs
@@ -327,18 +327,70 @@ impl WaitCell {
         }
     }
 
-    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    /// Asynchronously poll the given function `f` until a condition occurs,
+    /// using the [`WaitCell`] to only re-poll when notified.
     ///
     /// This can be used to implement a "wait loop", turning a "try" function
     /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
     /// "recv" or "send").
     ///
+    /// In particular, this function correctly *registers* interest in the [`WaitCell`]
+    /// prior to polling the function, ensuring that there is not a chance of a race
+    /// where the condition occurs AFTER checking but BEFORE registering interest
+    /// in the [`WaitCell`], which could lead to deadlock.
+    ///
+    /// This is intended to have similar behavior to `Condvar` in the standard library,
+    /// but asynchronous, and not requiring operating system intervention (or existence).
+    ///
+    /// In particular, this can be used in cases where interrupts or events are used
+    /// to signify readiness or completion of some task, such as the completion of a
+    /// DMA transfer, or reception of an ethernet frame. In cases like this, the interrupt
+    /// can wake the queue, allowing the polling function to check status fields for
+    /// partial progress or completion.
+    ///
     /// Consider using [`Self::wait_for_value()`] if your function does return a value.
+    ///
+    /// Consider using [`WaitQueue::wait_for()`] if you need multiple waiters.
     ///
     /// # Returns
     ///
     /// * [`Ok`]`(())` if the closure returns `true`.
     /// * [`Err`]`(`[`Closed`]`)` if the [`WaitCell`] is closed.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio::task;
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn test() {
+    /// use std::sync::Arc;
+    /// use maitake_sync::WaitCell;
+    /// use std::sync::atomic::{AtomicU8, Ordering};
+    ///
+    /// let queue = Arc::new(WaitCell::new());
+    /// let num = Arc::new(AtomicU8::new(0));
+    ///
+    /// let waiter = task::spawn({
+    ///     // clone items to move into the spawned task
+    ///     let queue = queue.clone();
+    ///     let num = num.clone();
+    ///     async move {
+    ///         queue.wait_for(|| num.load(Ordering::Relaxed) > 5).await;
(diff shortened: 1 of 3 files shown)


maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
...
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for_value()`] if your function does return a value.
+    ///
+    /// * Returns `Ok(())` if the closure returns `true`.
+    /// * Returns `Err(Closed)` if the [`WaitCell`] is closed.
```
hawkw, line 339: tiny nit: elsewhere, the `maitake_sync` docs forrmat return value documentation as a `# Returns` heading --- e.g., in this module, see [here](https://github.com/hawkw/mycelium/blob/e456145e84ad6339752251318e6cf5d296b65d51/maitake-sync/src/wait_cell.rs#L114-L125):

```suggestion
    /// # Returns
    ///
    /// * [`Ok`]`(())` if the closure returns `true`.
    /// * [`Err`]`(`[`Closed`]`)` if the [`WaitCell`] is closed.
```

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
...
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for()`] if your function does not return a value.
+    ///
+    /// * Returns `Ok(T)` if the closure returns `Some(T)`.
+    /// * Returns `Err(Closed)` if the [`WaitCell`] is closed.
```
hawkw, line 359: same tiny nit as above:
```suggestion
    /// # Returns
    ///
    /// * [`Ok`]`(T)` if the closure returns [`Some`]`(T)`.
    /// * [`Err`]`(`[`Closed`]`)` if the [`WaitCell`] is closed.
```

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
         }
     }
 
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
```
hawkw, line 330: Personally, I don't love this wording --- I think it would be more accurate to say that we are polling the _condition_, not the `WaitCell`. This function polls for the condition and uses the waitcell to _wait_ if that condition is not yet met.

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
...
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
```
hawkw, line 335: if you have the time, it would be nice to have an example here, but no biggie --- i'm fine with merging this without one.

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
...
+        loop {
+            let wait = self.subscribe().await;
+            if f() {
+                return Ok(());
+            }
+            wait.await?;
+        }
+    }
+
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
```
hawkw, line 350: similarly, I might adjust the wording here a bit...

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,46 @@ impl WaitCell {
         }
     }
 
+    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
```
hawkw, line 334: another use case is waiting for an interrupt handler to set a bit or something --- there's actually a lot of code in some of the mnemOS drivers that are basically manual implementations of this pattern, e.g. https://github.com/tosc-rs/mnemos/blob/45457b0e72d1bfb755d700f982964092fc116512/platforms/esp32c3-buddy/src/drivers/usb_serial.rs#L96-L104. it might be good to mention something like that here?

maitake-sync/src/wait_cell.rs
hawkw: Also, it might be good for these docs to mention that this method ensures that the future is registered to be woken up by the `WaitCell` *before* checking the condition, so that there isn't a potential race where a wakeup is missed while the condition is being checked, since that's...basically the whole point of having this method :)

It might even be worth comparing to `std::sync::Condvar`, which uses a Mutex to make similar guarantees?

maitake-sync/src/wait_queue.rs
hawkw, line 1: most of the docs feedback I left on the `WaitCell` code also applies here.

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,56 @@ impl WaitQueue {
...
+    /// Asynchronously poll the [`WaitQueue`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for_value()`] if your function does return a value.
+    ///
+    /// * Returns `Ok(())` if the closure returns `true`.
+    /// * Returns `Err(Closed)` if the [`WaitQueue`] is closed.
```
hawkw, line 696:
```suggestion
    /// # Returns
    ///
    /// * [`Ok`]`(())` if the closure returns `true`.
    /// * [`Err`]`(`[`Closed`]`)` if the [`WaitQueue`] is closed.
```

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,56 @@ impl WaitQueue {
...
+    /// Asynchronously poll the [`WaitQueue`] until a condition occurs
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// Consider using [`Self::wait_for()`] if your function does not return a value.
+    ///
+    /// * Returns `Ok(T)` if the closure returns `Some(T)`.
+    /// * Returns `Err(Closed)` if the [`WaitQueue`] is closed.
```
hawkw, line 721:
```suggestion
    /// * [`Ok`]`(T)` if the closure returns [`Some`]`(T)`.
    /// * [`Err`]`(`[`Closed`]`)` if the [`WaitQueue`] is closed.
```

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,56 @@ impl WaitQueue {
...
+    /// * Returns `Ok(())` if the closure returns `true`.
+    /// * Returns `Err(Closed)` if the [`WaitQueue`] is closed.
+    pub async fn wait_for<F: FnMut() -> bool>(&self, mut f: F) -> WaitResult<()> {
+        loop {
+            let wait = self.wait();
+            let mut pwait = core::pin::pin!(wait);
+            match pwait.as_mut().subscribe() {
+                Poll::Ready(wr) => wr?,
+                Poll::Pending => {}
+            }
```
hawkw, line 704: nit, take it or leave it: I think this could be written as 
```suggestion
            let _ = pwait.as_mut().subscribe()?;
```
since `Poll<Result<T, E>>` [implements `Try<Output = Poll<T>>`](https://doc.rust-lang.org/stable/std/task/enum.Poll.html#impl-Try-for-Poll%3CResult%3CT,+E%3E%3E) and then we can just throw out the `Poll` value.
 
on the other hand, maybe the match is clearer to the reader? i could be convinced either way.

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,56 @@ impl WaitQueue {
...
+            if f() {
+                return Ok(());
+            }
+            pwait.await?;
```
hawkw, line 708: nit, take it or leave it: i would personally just shadow the name `wait` instead of using `pwait`, 
```suggestion
            let mut wait = core::pin::pin!(wait);
            match wait.as_mut().subscribe() {
                Poll::Ready(wr) => wr?,
                Poll::Pending => {}
            }
            if f() {
                return Ok(());
            }
            wait.await?;
```
but, not a big deal either way.

hawkw (approved):
This looks great, thank you! I left some feedback, but it's mostly just me being pedantic about docs, plus some style suggestions that you're welcome to take if you like them, or to ignore if you don't.

It might be nice to have tests and/or examples for the new code, but it's simple enough that I'm happy to merge this regardless. We may want to add tests asserting that the futures returned by these functions are `Send`/`Sync` when the input closure is `Send`/`Sync`, though, so that future internal changes don't introduce an accidental API compatibility breakage.

jamesmunns:
github was being weird about applying suggestions, not sure why it thought they were overlapping. I plan to come back to add docs and address the remaining open topics

hawkw:
Great, I'll be happy to merge this whenever the docs are ready (and ideally, when we can get a clean CI build on `main` 😬)

maitake-sync/src/wait_cell.rs
```diff
@@ -327,18 +327,70 @@ impl WaitCell {
...
 
-    /// Asynchronously poll the [`WaitCell`] until a condition occurs
+    /// Asynchronously poll the given function `f` until a condition occurs,
+    /// using the [`WaitCell`] to only re-poll when notified.
     ///
     /// This can be used to implement a "wait loop", turning a "try" function
     /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
     /// "recv" or "send").
     ///
+    /// In particular, this function correctly *registers* interest in the [`WaitCell`]
```
hawkw, line 337: very tiny nit: maybe worth linking to the `subscribe` method here where you say "registers"?

maitake-sync/src/wait_cell.rs
```diff
@@ -327,6 +327,159 @@ impl WaitCell {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio::task;
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn test() {
+    /// use std::sync::Arc;
+    /// use maitake_sync::WaitCell;
+    /// use std::sync::atomic::{AtomicU8, Ordering};
```
hawkw, line 440: nit, take it or leave it: maybe the examples could get these from alloc/core just so that they "look more like" no-std code?
```suggestion
    /// use alloc::sync::Arc;
    /// use maitake_sync::WaitCell;
    /// use core::sync::atomic::{AtomicU8, Ordering};
```

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,195 @@ impl WaitQueue {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio::task;
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn test() {
+    /// use std::sync::Arc;
+    /// use maitake_sync::WaitQueue;
+    /// use std::sync::atomic::{AtomicU8, Ordering};
```
hawkw, line 725: similarly, maybe import these from core?

maitake-sync/src/wait_queue.rs
```diff
@@ -684,6 +684,195 @@ impl WaitQueue {
...
+    /// # test();
+    /// ```
+    pub async fn wait_for_value<T, F: FnMut() -> Option<T>>(&self, mut f: F) -> WaitResult<T> {
+        loop {
+            let wait = self.wait();
+            let mut wait = core::pin::pin!(wait);
+            match wait.as_mut().subscribe() {
+                Poll::Ready(wr) => wr?,
+                Poll::Pending => {}
+            }
```
hawkw, line 868: nit, take it or leave it: could be one line shorter as:
```suggestion
            if let Poll::Ready(wr) = wait.as_mut().subscribe() {
                wr?
            }
```

hawkw (approved):
docs and tests look good to me --- happy to merge this once CI is green!

maitake-sync/src/wait_cell.rs
jamesmunns: I could, but it's through the `Wait` struct, and I didn't want to be distracting. If you want it, I can add it.

maitake-sync/src/wait_cell.rs
jamesmunns: I can do this, but other tests in the same module are already using std, so I went with consistency. I can switch them all over if you'd like.

maitake-sync/src/wait_queue.rs
jamesmunns: will probably leave for clarity

jamesmunns:
@hawkw tests are green (🎉), let me know what you think re: `std` and `subscribe`.

maitake-sync/src/wait_cell.rs
hawkw: Oh. Well, in that case, never mind!

maitake-sync/src/wait_cell.rs
hawkw: not a big deal to me either way, this is fine!

PR #481, opened 2024-07-20
feat(maitake-sync): rename `EnqueueWait` to `Subscribe`

hawkw:
`maitake-sync`'s `waitmap::Wait` future has an `enqueue` method that
does the same thing as the `subscribe` methods on `WaitCell` and
`WaitQueue`...but has a different name. I think this might be due to
`enqueue` being added earlier or something, or maybe we just weren't
paying attention. Regardless, I figured it was nicer to use the same
name for the method that does the same thing on these different types.

This commit renames the `waitmap::Wait::enqueue` method to `subscribe`,
and the future it returns from `EnqueueWait` to `Subscribe`, analogously
to `WaitQueue` and `WaitCell`. I've left a deprecated type alias and a
deprecated function alias in place, so this isn't a breaking change:
users who reference `Wait::enqueue` or `EnqueueWait` will still be able
to compile their code, but they'll get a warning.

cc @jamesmunns, since I believe you added this API?

### DIFF
--- a/maitake-sync/src/wait_map.rs
+++ b/maitake-sync/src/wait_map.rs
@@ -267,7 +267,7 @@ impl<'map, 'wait, K: PartialEq, V> Wait<'map, K, V> {
     ///     // first pin the `wait` future, to ensure that it does not move
     ///     // until it has been completed.
     ///     pin_mut!(wait);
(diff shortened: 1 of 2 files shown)


jamesmunns (approved):
hell yeah

Issue #393, opened 2023-01-11
Compiling error on mach-o 


nicolasdanelon:
I'm having the following error while I run `cargo run-x64`:

```
   Compiling owo-colors v2.1.0
LLVM ERROR: Global variable '_ZN8mycotest1_4TEST17h5b175f4f09402ec8E' has an invalid section specifier 'MyceliumTests': mach-o section specifier requires a segment and section separated by a comma.
error: could not compile `mycotest`
warning: build failed, waiting for other jobs to finish...
```

- rustc --version `>>` rustc 1.68.0-nightly (bdb07a8ec 2022-12-11)
- OS: Ventura 13.1 `>>` Darwin sov.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:04:44 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8103 arm64

hawkw:
Huh, that's interesting. We _should_ be compiling Mycelium as an ELF rather than a mach-o regardless of the host OS that's building the Mycelium binary, so this is a pretty surprising error.  It seems like we're probably missing some configuration somewhere and the macOS toolchain is trying to build a mach-o object.

PR #484, opened 2024-08-10
change custom link section to support apple mach-o format

spookyvision:


hawkw (approved):
It feels quite weird to me that we have to do this, since this code *should* never actually be built for a MacOS target: the `decl_test!` macro is only used in code that goes into the Mycelium kernel, which should always be built with `target_os = "none"`. 

So, perhaps we could simplify this to just:
```rust
#[cfg_attr(target_os = "none", link_section = "MyceliumTests")]
```
to avoid emitting the attribute at all when not building a Mycelium kernel image?

With that said, if the current code fixes the build on MacOS dev machines, I'm fine with it either way!

BTW, I believe this fixes issue #393, so let's add that to the commit message before merging!

hawkw:
> So, perhaps we could simplify this to just:
> 
> ```rust
> #[cfg_attr(target_os = "none", link_section = "MyceliumTests")]
> ```
> 
> to avoid emitting the attribute at all when not building a Mycelium kernel image?

I don't actually know whether this will work or not, FWIW. If the macOS linker is involved even when linking cross-compiled builds for `x86_64-unknown-none`, we might have to placate it like this anyway? Up to you if you want to investigate whether we can get away with just not having the attribute.

PR #499, opened 2024-12-31
make local apic actually local

hawkw:


hawkw:
for the record, i knew that putting tracing in the PIT sleep functions would trash their accuracy when i wrote that code, but i did it anyway because we weren't actually using it for calibration at the time...

PR #514, opened 2025-02-01
fix(maitake-sync): for cs-mutex, enter cs before locking mutex

### LINKED ISSUE #513: Soundness of CriticalSectionDefaultMutex implementation
I am currently experiencing some deadlocks with the `WaitQueue` when using the cs-backed mutex in a single-core system with interrupts. The stack trace seems to be stuck at the `SpinLock` inside the `CriticalSectionDefaultMutex`, which in my mind is a place the cs-mutex should never be on a single-core system.

I looked at the code:

https://github.com/hawkw/mycelium/blob/main/maitake-sync/src/blocking/default_mutex.rs#L294-L326

Specifically the `ScopedRawMutex` trait function

```rust
fn with_lock<R>(&self, f: impl FnOnce() -> R) -> R {
    self.0.with_lock(|| critical_section::with(|_cs| f()))
}
```
with the `SpinLock::with_lock` being
```rust
fn with_lock<R>(&self, f: impl FnOnce() -> R) -> R {
    self.lock();
    // Using a drop guard ensures that the mutex is unlocked when this
    // function exits, even if `f()` panics.
    let _unlock = Unlock(self);
    f()
}
``` 

Is it not incorrect to first lock the spin-lock, and then enter the critical section after? If an interrupt fires between locking the `SpinLock` and entering the critical section with `f()` the interrupt context would just get stuck in the spin loop trying to lock.

peterkrull:
Closes #513

### DIFF
--- a/maitake-sync/src/blocking/default_mutex.rs
+++ b/maitake-sync/src/blocking/default_mutex.rs
@@ -309,13 +309,13 @@ mod cs_impl {
         #[track_caller]
         #[inline(always)]
         fn with_lock<R>(&self, f: impl FnOnce() -> R) -> R {
-            self.0.with_lock(|| critical_section::with(|_cs| f()))
+            critical_section::with(|_cs| self.0.with_lock(f))
         }
 
         #[track_caller]
         #[inline(always)]
         fn try_with_lock<R>(&self, f: impl FnOnce() -> R) -> Option<R> {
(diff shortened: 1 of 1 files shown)


hawkw (approved):

hawkw:
Thanks, this looks good to me! I'll merge this and your other change #515, and get a release out shortly!

PR #515, opened 2025-02-01
fix(maitake-sync): add proper `ScopedRawMutex` bound to `wait_map::Wait`

peterkrull:
I was unable to await the `Wait` when using a mutex defined in the `mutex` crate because a trait bound was not met. The current trait impl is not general enough.

### DIFF
--- a/maitake-sync/src/wait_map.rs
+++ b/maitake-sync/src/wait_map.rs
@@ -997,7 +997,11 @@ unsafe impl<K: PartialEq, V> Linked<list::Links<Waiter<K, V>>> for Waiter<K, V> 
 
 // === impl Wait ===
 
-impl<K: PartialEq, V> Future for Wait<'_, K, V> {
+impl<K, V, Lock> Future for Wait<'_, K, V, Lock>
+where
+    K: PartialEq,
+    Lock: ScopedRawMutex,
+{
     type Output = WaitResult<V>;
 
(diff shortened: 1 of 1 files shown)


peterkrull:
This seems to also be a thing for other types which specifically use the DefaultMutex default type, like the `Lock` for the async mutex.
```rust
pub struct Lock<'a, T: ?Sized, L: ScopedRawMutex = DefaultMutex> { .. }

impl<'a, T> Future for Lock<'a, T> {..}
```

hawkw (approved):
Whoops! Thanks for fixing this!

We should probably add some tests that actually construct each synchronization primitive with an overridden `ScopedRawMutex` type and check that all the expected trait impls are present, to prevent against future regressions with this sort of thing. I don't think it's worth holding up merging this fix for such tests, but I'll open an issue because I think it's worth doing.

hawkw:
Opened #516 for adding tests to ensure expected `Future` impls are present.

hawkw:
> This seems to also be a thing for other types which specifically use the DefaultMutex default type, like the `Lock` for the async mutex.
> 
> ```rust
> pub struct Lock<'a, T: ?Sized, L: ScopedRawMutex = DefaultMutex> { .. }
> 
> impl<'a, T> Future for Lock<'a, T> {..}
> ```

#517 fixes this for the `mutex::Lock` future. Checking the other async synchronization primitives (`WaitQueue`, `RwLock`, and `Semaphore`), it looks like their `Future` types already have proper trait bounds, but I'll add regression tests for them as well.

PR #518, opened 2025-02-06
test(maitake-sync): test futures impl `Future`

### LINKED ISSUE #516: maitake-sync: test sync type `Future` impls (etc) with overridden `ScopedRawMutex`
See #515, where `wait_map`'s `Wait` future's `Future` impl only exists with `DefaultMutex`.

> We should probably add some tests that actually construct each synchronization primitive with an overridden `ScopedRawMutex` type and check that all the expected trait impls are present, to prevent against future regressions with this sort of thing. I don't think it's worth holding up merging this fix for such tests, but I'll open an issue because I think it's worth doing.
>
> _Originally posted by @hawkw in https://github.com/hawkw/mycelium/pull/515#pullrequestreview-2599435497_

hawkw:
Depends on #517

As described in #516, this commit adds tests ensuring that `Future` impls are always present for sync primitive future types when using user-provided `ScopedRawMutex`/`RawMutex` impls. This should help protect against future regressions.

Fixes #516

### DIFF
--- a/maitake-sync/src/mutex.rs
+++ b/maitake-sync/src/mutex.rs
@@ -348,8 +348,12 @@ where
 
 // === impl Lock ===
 
-impl<'a, T> Future for Lock<'a, T> {
-    type Output = MutexGuard<'a, T>;
+impl<'a, T, L> Future for Lock<'a, T, L>
+where
+    T: ?Sized,
+    L: ScopedRawMutex,
+{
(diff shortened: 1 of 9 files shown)


hawkw:
Oh, ugh, some of these tests don't actually work, as they use APIs where creating the future actually _does_ attempt to lock immediately. I'll rework it to be based on the types instead.

PR #520, opened 2025-03-24
feat(cordyceps): singly-linked intrusive `SortedList`

jamesmunns:
This PR implements:

* A new method for `TransferStack` that reports whether the list was previously empty
* A new type, `SortedList`, that is a sorted singly-linked-list

These are changes I am making for experimentation with embassy's scheduler, which I am using `cordyceps` for.

These changes are currently functional and minimally tested, but probably could use tweaks to better fit naturally in the library.

### DIFF
--- a/cordyceps/src/sorted_list.rs
+++ b/cordyceps/src/sorted_list.rs
@@ -6,7 +6,12 @@
 //! [Intrusive]: crate#intrusive-data-structures
 
 use crate::{Linked, Stack};
-use core::{cmp::{Ord, Ordering}, marker::PhantomData, ptr::NonNull};
+use core::{
+    cmp::{Ord, Ordering},
+    fmt,
+    marker::PhantomData,
+    ptr::NonNull,
+};
 
 pub use crate::stack::Links;
 
@@ -24,7 +29,6 @@ pub use crate::stack::Links;
 ///
 /// If your type `T` does NOT implement [`Ord`], or you want to use
 /// a custom sorting anyway, consider using [`SortedList::new_with_cmp()`]
-#[derive(Debug)]
 pub struct SortedList<T: Linked<Links<T>>> {
     head: Option<NonNull<T>>,
     // Returns if LHS is less/same/greater than RHS
@@ -189,7 +193,7 @@ impl<T: Linked<Links<T>>> SortedList<T> {
         // * insert between cursor and cursor.next (if elem is < c.next)
         // * just iterate (if elem >= c.next)
         loop {
-            // SAFETY: We have exclusive access to the list, we are allowed to
+            // Safety: We have exclusive access to the list, we are allowed to
             // access and mutate it (carefully)
             unsafe {
                 // Get the cursor's links
@@ -272,6 +276,16 @@ impl<T: Linked<Links<T>>> Extend<T::Handle> for SortedList<T> {
     }
 }
 
+impl<T> fmt::Debug for SortedList<T>
+where
+    T: Linked<Links<T>>,
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let Self { head, func: _ } = self;
+        f.debug_struct("SortedList").field("head", head).finish()
+    }
+}
+
 impl<T: Linked<Links<T>>> Drop for SortedList<T> {
     fn drop(&mut self) {
         // We just turn the list into a stack then run the stack drop code.
@@ -285,7 +299,6 @@ impl<T: Linked<Links<T>>> Drop for SortedList<T> {
 }
 
 /// A borrowing iterator of a [`SortedList`]
-#[derive(Debug)]
 pub struct SortedListIter<'a, T: Linked<Links<T>>> {
     _plt: PhantomData<&'a SortedList<T>>,
     node: Option<NonNull<T>>,
--- a/cordyceps/src/sorted_list.rs
+++ b/cordyceps/src/sorted_list.rs
@@ -23,7 +23,7 @@ pub use crate::stack::Links;
 /// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
 ///
 /// If your type `T` does NOT implement [`Ord`], or you want to use
-/// a custom sorting anyway, consider using [`SortedList::new_custom()`]
+/// a custom sorting anyway, consider using [`SortedList::new_with_cmp()`]
 #[derive(Debug)]
 pub struct SortedList<T: Linked<Links<T>>> {
     head: Option<NonNull<T>>,
@@ -45,39 +45,39 @@ where
     /// Create a new (empty) sorted list, sorted LEAST FIRST
     ///
     /// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
-    /// * Consider using [`SortedList::new_custom()`] if you want to provide your own sorting
+    /// * Consider using [`SortedList::new_with_cmp()`] if you want to provide your own sorting
     ///   implementation.
     ///
     /// If two items are considered of equal value, new values will be placed AFTER
     /// old values.
     #[must_use]
     pub const fn new_min() -> Self {
-        Self::new_custom(T::cmp)
+        Self::new_with_cmp(T::cmp)
     }
 
     /// Create a new sorted list, consuming the stack, sorted LEAST FIRST
     #[must_use]
     pub fn from_stack_min(stack: Stack<T>) -> Self {
-        Self::from_stack_custom(stack, T::cmp)
+        Self::from_stack_with_cmp(stack, T::cmp)
     }
 
     /// Create a new (empty) sorted list, sorted GREATEST FIRST
     ///
     /// * Consider using [`SortedList::new_min()`] if you want **smallest** items sorted first.
-    /// * Consider using [`SortedList::new_custom()`] if you want to provide your own sorting
+    /// * Consider using [`SortedList::new_with_cmp()`] if you want to provide your own sorting
     ///   implementation.
     ///
     /// If two items are considered of equal value, new values will be placed AFTER
     /// old values.
     #[must_use]
     pub const fn new_max() -> Self {
-        Self::new_custom(invert_sort::<T>)
+        Self::new_with_cmp(invert_sort::<T>)
     }
 
     /// Create a new sorted list, consuming the stack, sorted GREATEST FIRST
     #[must_use]
     pub fn from_stack_max(stack: Stack<T>) -> Self {
-        Self::from_stack_custom(stack, invert_sort::<T>)
+        Self::from_stack_with_cmp(stack, invert_sort::<T>)
     }
 }
 
@@ -104,7 +104,7 @@ impl<T: Linked<Links<T>>> SortedList<T> {
     ///
     /// If two items are considered of equal value, new values will be placed AFTER
     /// old values.
-    pub const fn new_custom(f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+    pub const fn new_with_cmp(f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
         Self {
             func: f,
             head: None,
@@ -112,8 +112,8 @@ impl<T: Linked<Links<T>>> SortedList<T> {
     }
 
     /// Create a new sorted list, consuming the stack, using the provided ordering function
-    pub fn from_stack_custom(stack: Stack<T>, f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
-        let mut slist = Self::new_custom(f);
+    pub fn from_stack_with_cmp(stack: Stack<T>, f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        let mut slist = Self::new_with_cmp(f);
         slist.extend(stack);
         slist
     }
@@ -414,7 +414,7 @@ mod loom {
     #[test]
     fn slist_basic() {
         loom::model(|| {
-            let mut slist = SortedList::<Entry>::new_custom(|lhs, rhs| lhs.val.cmp(&rhs.val));
+            let mut slist = SortedList::<Entry>::new_with_cmp(|lhs, rhs| lhs.val.cmp(&rhs.val));
             // Insert out of order
             slist.insert(Entry::new(20));
             slist.insert(Entry::new(10));
--- a/cordyceps/src/sorted_list.rs
+++ b/cordyceps/src/sorted_list.rs
@@ -1,0 +1,298 @@
+//! [Intrusive], singly-linked, sorted, linked-list.
+//!
+//! See the documentation for the [`SortedList`] and [`SortedListIter`] types for
+//! details.
+//!
+//! [intrusive]: crate#intrusive-data-structures
+
+use crate::{Linked, Stack};
+use core::{marker::PhantomData, ptr::NonNull};
+
+pub use crate::stack::Links;
+
+/// A sorted singly linked list
+///
+/// This behaves similar to [`Stack`], in that it is a singly linked list,
+/// however items are stored in an ordered fashion. This means that insertion
+/// is an O(n) operation, and retrieval of the first item is an O(1) operation.
+///
+/// It allows for a user selected ordering operation, see [`SortedList::new`]
+/// for usage example.
+#[derive(Debug)]
+pub struct SortedList<T: Linked<Links<T>>> {
+    head: Option<NonNull<T>>,
+    // Returns if LHS is less/same/greater than RHS
+    func: fn(&T, &T) -> core::cmp::Ordering,
+}
+
+impl<T: Linked<Links<T>>> SortedList<T> {
+    /// Create a new (empty) sorted list with the given ordering function
+    ///
+    /// If `T` contained an `i32`, and you wanted the SMALLEST items at the
+    /// front, then you could provide a function something like:
+    ///
+    /// ```rust
+    /// let f: fn(&i32, &i32) -> core::cmp::Ordering = |lhs, rhs| {
+    ///     lhs.cmp(rhs)
+    /// };
+    /// ```
+    ///
+    /// SortedList takes a function (and not just [Ord]) so you can use it on types
+    /// that don't have a general way of ordering them, allowing you to select a
+    /// specific metric within the sorting function.
+    ///
+    /// If two items are considered of equal value, new values will be placed AFTER
+    /// old values.
+    pub const fn new(f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        Self {
+            func: f,
+            head: None,
+        }
+    }
+
+    /// Create a new sorted list, consuming the stack, using the provided ordering function
+    pub fn from_stack(stack: Stack<T>, f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        let mut slist = Self::new(f);
+        slist.extend(stack);
+        slist
+    }
+
+    /// Pop the front-most item from the list, returning it by ownership (if it exists)
+    #[must_use]
+    pub fn pop_front(&mut self) -> Option<T::Handle> {
+        test_trace!(?self.head, "SortedList::pop_front");
+        let head = self.head.take()?;
+        unsafe {
+            // Safety: we have exclusive ownership over this chunk of stack.
+
+            // advance the head link to the next node after the current one (if
+            // there is one).
+            self.head = T::links(head).as_mut().next.with_mut(|next| (*next).take());
+
+            test_trace!(?self.head, "SortedList::pop -> popped");
+
+            // return the current node
+            Some(T::from_ptr(head))
+        }
+    }
+
+    /// Insert a single item into the list, in it's sorted order position
+    pub fn insert(&mut self, element: T::Handle) {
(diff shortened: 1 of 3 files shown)


jamesmunns:
These changes are exercised in https://github.com/jamesmunns/embassy/pull/1 as well.

cordyceps/src/stack.rs
```diff
@@ -400,6 +416,245 @@ impl<T> Default for Links<T> {
...
+    /// specific metric within the sorting function.
+    ///
+    /// If two items are considered of equal value, new values will be placed AFTER
+    /// old values.
+    pub const fn new(f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        Self {
+            func: f,
+            head: None,
+        }
+    }
```
hawkw, line 462: It feels to me that the common case is to use `cmp::Ord` for this, and to have an "override" constructor that takes a custom comparator? I might consider changing the constructors around a bit to look like this:

```rust
impl<T> SortedList<T>
where
    T: Linked<Links<T>>,
    T: core::cmp::Ord,
{
    pub const fn new() -> Self {
        Self::with_cmp(T::cmp)
    }

    pub fn from_stack(stack: Stack<T>) -> Self {
        Self::from_stack_with_cmp(stack, T::cmp)
    }
}

impl<T> Default for SortedList<T>
where
    T: Linked<Links<T>>,
    T: core::cmp::Ord,
{
    fn default() -> Self {
        Self::new()
    }
}

impl<T: Linked<Links<T>>> SortedList<T> {
    pub const fn with_cmp(cmp: fn(&T, &T) -> core::cmp::Ordering) -> Self {
        Self {
            func: cmp,
            head: None,
        }
    }

    pub fn from_stack_with_cmp(stack: Stack<T>) -> Self {
        let mut slist = Self::new(f);
        slist.extend(stack);
        slist
    }

    // ... put other methods here ...
}

```

What do you think?

cordyceps/src/stack.rs
```diff
@@ -400,6 +416,245 @@ impl<T> Default for Links<T> {
...
+            self.insert(elem);
+        }
+    }
+}
+
+impl<T: Linked<Links<T>>> Drop for SortedList<T> {
+    fn drop(&mut self) {
+        // TODO: This is a hack. We just turn the list into a stack then run the
+        // stack drop code. It already has correct + tested logic for dropping
+        // a singly linked list of items one at a time.
```
hawkw, line 625: I actually think this is the Correct Thing To Do Here, so I'm unsure if it really needs a TODO comment.

cordyceps/src/stack.rs
```diff
@@ -400,6 +416,245 @@ impl<T> Default for Links<T> {
...
+            self.head = T::links(head).as_mut().next.with_mut(|next| (*next).take());
+
+            test_trace!(?self.head, "SortedList::pop -> popped");
+
+            // return the current node
+            Some(T::from_ptr(head))
+        }
+    }
+
+    /// Insert a single item into the list, in it's sorted order position
```
hawkw, line 490: Should probably note that this is _O_(n)?

cordyceps/src/stack.rs
```diff
@@ -400,6 +416,245 @@ impl<T> Default for Links<T> {
     }
 }
 
+// --------------------------------------------------------------------------
+// This should live somewhere else, but for now it doesn't.
```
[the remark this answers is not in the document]
hawkw: I'd probably stick it in its own top-level `pub mod sorted_list` --- feel free to make any `Stack`/`TransferStack` bits `pub(crate)` as necessary.

cordyceps/src/stack.rs
jamesmunns: That seems reasonable! My main use case is a "weird" one for embassy (or similar executors) where it might not totally make sense to make `TaskHeader` impl `Ord` (especially conditionally). That being said, if people want to use it for normal-sorted-list-activities, I think having a helper like this makes sense!

cordyceps/src/stack.rs
jamesmunns: Yep, will weaken. Didn't know if we wanted to have some common drop fn instead to make the intent more clear, but I agree it doesn't make sense to impl the functionality twice.

cordyceps/src/stack.rs
jamesmunns: Ooh, good call. I did note this in the place where I call this, so I'll steal that.

cordyceps/src/stack.rs
```diff
@@ -152,9 +168,9 @@ where
                 .head
                 .compare_exchange_weak(head, ptr.as_ptr(), AcqRel, Acquire)
             {
-                Ok(_) => {
+                Ok(old) => {
                     test_trace!(?ptr, ?head, "TransferStack::push -> pushed");
-                    return;
+                    return old.is_null();
```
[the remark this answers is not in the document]
jamesmunns: Also if you prefer that I don't just "check then throw away" for the main `push` method, I can do that. I'm hoping the optimizer is smart enough to just skip the check in that case, but I definitely haven't checked, and it's probably not guaranteed.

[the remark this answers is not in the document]
hawkw: I would certainly _hope_ LLVM figures that out, but it's hard to say...

jamesmunns:
I will likely not plan to add any Smarter impls for SortedList for now, but if you would like me to leave room for other impls (SkipSortedList?) somewhere, I can do so!

cordyceps/src/stack.rs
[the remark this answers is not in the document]
jamesmunns: Re-homed

cordyceps/src/stack.rs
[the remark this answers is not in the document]
jamesmunns: Added

jamesmunns:
@hawkw I think I've addresses all comments, back to you, let me know if I can help get this published!

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+pub struct SortedList<T: Linked<Links<T>>> {
+    head: Option<NonNull<T>>,
+    // Returns if LHS is less/same/greater than RHS
+    func: fn(&T, &T) -> core::cmp::Ordering,
+}
+
+#[inline]
+fn invert_sort<T: core::cmp::Ord>(a: &T, b: &T) -> core::cmp::Ordering {
+    // Inverted sort order!
+    T::cmp(b, a)
```
hawkw, line 37: I was wondering why `core:;cmp` didn't already have something like this baked in, and I guess that's because it has [this instead](https://doc.rust-lang.org/stable/std/cmp/struct.Reverse.html), which is a bit annoying to use here...

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// };
+    /// ```
+    ///
+    /// SortedList takes a function (and not just [Ord]) so you can use it on types
+    /// that don't have a general way of ordering them, allowing you to select a
+    /// specific metric within the sorting function.
+    ///
+    /// If two items are considered of equal value, new values will be placed AFTER
+    /// old values.
+    pub const fn new_custom(f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
```
hawkw, line 103: Naming nit, take it or leave it: I don't love the name "new custom", as it's not *super* obvious what's custom --- I might have called this `with_cmp` (or `new_with_cmp`) or something, to make it clearer what specifically is being customized?

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// If two items are considered of equal value, new values will be placed AFTER
+    /// old values.
+    pub const fn new_max() -> Self {
+        Self::new_custom(invert_sort::<T>)
+    }
+
+    /// Create a new sorted list, consuming the stack, sorted GREATEST FIRST
+    pub fn from_stack_max(stack: Stack<T>) -> Self {
+        Self::from_stack_custom(stack, invert_sort::<T>)
+    }
```
hawkw, line 77: teensy nit: i like to add `#[must_use]` to constructors so that you get a warning if you construct a collection and forget to actually use it:
```suggestion
    /// Create a new (empty) sorted list, sorted LEAST FIRST
    ///
    /// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
    /// * Consider using [`SortedList::new_custom()`] if you want to provide your own sorting
    ///   implementation.
    ///
    /// If two items are considered of equal value, new values will be placed AFTER
    /// old values.
    #[must_use]
    pub const fn new_min() -> Self {
        Self::new_custom(T::cmp)
    }

    /// Create a new sorted list, consuming the stack, sorted LEAST FIRST
    #[must_use]
    pub fn from_stack_min(stack: Stack<T>) -> Self {
        Self::from_stack_custom(stack, T::cmp)
    }

    /// Create a new (empty) sorted list, sorted GREATEST FIRST
    ///
    /// * Consider using [`SortedList::new_min()`] if you want **smallest** items sorted first.
    /// * Consider using [`SortedList::new_custom()`] if you want to provide your own sorting
    ///   implementation.
    ///
    /// If two items are considered of equal value, new values will be placed AFTER
    /// old values.
    #[must_use]
    pub const fn new_max() -> Self {
        Self::new_custom(invert_sort::<T>)
    }

    /// Create a new sorted list, consuming the stack, sorted GREATEST FIRST
    #[must_use]
    pub fn from_stack_max(stack: Stack<T>) -> Self {
        Self::from_stack_custom(stack, invert_sort::<T>)
    }
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// * Consider using [`SortedList::new_custom()`] if you want to provide your own sorting
+    ///   implementation.
+    ///
+    /// If two items are considered of equal value, new values will be placed AFTER
+    /// old values.
+    pub const fn new_max() -> Self {
+        Self::new_custom(invert_sort::<T>)
+    }
+
+    /// Create a new sorted list, consuming the stack, sorted GREATEST FIRST
```
hawkw, line 74: It would be nice if the docs for these had examples showing the min- and max-ordering behaviors...but, not a blocker for this PR if that's too much effort, we can add them later.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+use crate::{Linked, Stack};
+use core::{marker::PhantomData, ptr::NonNull};
+
+pub use crate::stack::Links;
+
+/// A sorted singly linked list
+///
+/// This behaves similar to [`Stack`], in that it is a singly linked list,
+/// however items are stored in an ordered fashion. This means that insertion
+/// is an `O(n)` operation, and retrieval of the first item is an `O(1)` operation.
```
hawkw, line 17: nit: i would format this as:
```suggestion
/// is an _O_(_n_) operation, and retrieval of the first item is an _O_(1) operation.
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+///
+/// It allows for a user selected ordering operation. If your type `T` implements
+/// [`Ord`]:
+///
+/// * Consider using [`SortedList::new_min()`] if you want **smallest** items sorted first.
+/// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
+///
+/// If your type `T` does NOT implement [`Ord`], or you want to use
+/// a custom sorting anyway, consider using [`SortedList::new_custom()`]
```
hawkw, line 26: Let's also add a note about the fact that elements in a `SortedList` must implement  `Linked<stack::Links<T>>`, and --- critically --- that this means that the SAME links are reused when an item is linked into the sorted list and when it's linked into a stack. That means that you can link the same element into either structure, but you can't have something that's linked into a sorted list and a stack at the same time (without wrapper structs that have separate sets of links, which is a thing we do in maitake a bit...)

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
+//! [Intrusive], singly-linked, sorted, linked-list.
+//!
+//! See the documentation for the [`SortedList`] and [`SortedListIter`] types for
+//! details.
+//!
+//! [intrusive]: crate#intrusive-data-structures
```
hawkw, line 6: nit: i _think_ this link is broken, because you capitalized it above...
```suggestion
//! [Intrusive]: crate#intrusive-data-structures
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
+//! [Intrusive], singly-linked, sorted, linked-list.
```
hawkw, line 1: docs nit:
```suggestion
//! [Intrusive], singly-linked, sorted, linked list.
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// Create a new sorted list, consuming the stack, using the provided ordering function
+    pub fn from_stack_custom(stack: Stack<T>, f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        let mut slist = Self::new_custom(f);
+        slist.extend(stack);
+        slist
+    }
+
+    /// Pop the front-most item from the list, returning it by ownership (if it exists)
+    ///
+    /// This is an `O(1)` operation.
```
hawkw, line 119: nit:
```suggestion
    /// This is an _O_(1) operation.
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    }
+
+    /// Create a new sorted list, consuming the stack, using the provided ordering function
+    pub fn from_stack_custom(stack: Stack<T>, f: fn(&T, &T) -> core::cmp::Ordering) -> Self {
+        let mut slist = Self::new_custom(f);
+        slist.extend(stack);
+        slist
+    }
+
+    /// Pop the front-most item from the list, returning it by ownership (if it exists)
```
hawkw, line 117: it would be nice if this explicitly stated somewhere that "front-most" means based on the ordering, not based on insertion order... part of me wants to call this `pop_min` or `pop_least`, but the fact that we have a reversed-ordering constructor makes that confusing...perhaps if we did away with that and made the constructors always be min-ordered, and said "use `core::cmp::Reverse` as a wrapper if you want max-first"...but that feels more awkward...

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    ///
```
hawkw, line 118: not going to block on this, but an example would be nice

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
+    ///
+    /// If `T` contained an `i32`, and you wanted the SMALLEST items at the
+    /// front, then you could provide a function something like:
+    ///
+    /// ```rust
+    /// let f: fn(&i32, &i32) -> core::cmp::Ordering = |lhs, rhs| {
+    ///     lhs.cmp(rhs)
+    /// };
+    /// ```
```
hawkw, line 95: it would be nice if this was a full-on example showing the use of this constructor as well, although i know writing examples for `cordyceps` APIs is a bit annoying since you have to implement the whole `Linked` trait and such. i'm fine with not doing that now if it's too anoying.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+            test_trace!(?self.head, "SortedList::pop -> popped");
+
+            // return the current node
+            Some(T::from_ptr(head))
+        }
+    }
+
+    /// Insert a single item into the list, in it's sorted order position
+    ///
+    /// This is an `O(n)` operation.
```
hawkw, line 140: nit:
```suggestion
    /// This is an _O_(_n_) operation.
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+            self.head = T::links(head).as_mut().next.with_mut(|next| (*next).take());
+
+            test_trace!(?self.head, "SortedList::pop -> popped");
+
+            // return the current node
+            Some(T::from_ptr(head))
+        }
+    }
+
+    /// Insert a single item into the list, in it's sorted order position
```
hawkw, line 138: nit:
```suggestion
    /// Insert a single item into the list, in its sorted order position
```

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+    ///
```
hawkw, line 139: it would be nice if we mentioned the behavior where two items are equal in this doc comment, too.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+        // If the head is null, then just place the item
+        let Some(mut cursor) = self.head else {
+            // todo: assert element.next is null?
+            self.head = Some(ptr);
+            return;
+        };
+
+        // Special case for head: do we replace current head with new element?
+        {
+            // compare, but make sure we drop the live reference to the cursor
```
hawkw, line 159: is this required for a safety invariant? it would be nice to state why, if so.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+            let cmp = {
+                let cref = unsafe { cursor.as_ref() };
+                (self.func)(cref, eref)
+            };
+
+            // If cursor node is LESS or EQUAL: keep moving.
+            // If cursor node is GREATER: we need to place the new item BEFORE
+            if cmp == core::cmp::Ordering::Greater {
```
hawkw, line 167: nit, take it or leave it: maybe we should just import `core::cmp::Ordering` in this file...

hawkw (approved):
Looks good to me overall! I had some minor nits, mostly about the comments, but this should be good to merge once those are addressed.

Would you mind changing the commit message to something like 
> feat(cordyceps): singly-linked intrusive `SortedList` 

or something before merging? that way, it should look nice when generating the changelog...

cordyceps/src/sorted_list.rs
jamesmunns: Ah interesting! I should have gone looking, but yeah I think the current solution is reasonable.

cordyceps/src/sorted_list.rs
jamesmunns: Sure! I can do that. I'll probably go `new_with_cmp`.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+///
+/// It allows for a user selected ordering operation. If your type `T` implements
+/// [`Ord`]:
+///
+/// * Consider using [`SortedList::new_min()`] if you want **smallest** items sorted first.
+/// * Consider using [`SortedList::new_max()`] if you want **largest** items sorted first.
+///
+/// If your type `T` does NOT implement [`Ord`], or you want to use
+/// a custom sorting anyway, consider using [`SortedList::new_custom()`]
+#[derive(Debug)]
```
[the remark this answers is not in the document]
jamesmunns: I'll make it consistent with whatever `Stack` currently does!

cordyceps/src/sorted_list.rs
jamesmunns: Same note about examples

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+        test_trace!(?ptr, ?self.head, "SortedList::insert");
+
+        // Take a long-lived reference to the new element
+        let eref = unsafe { ptr.as_ref() };
+
+        // Special case for empty head
+        //
+        // If the head is null, then just place the item
+        let Some(mut cursor) = self.head else {
+            // todo: assert element.next is null?
```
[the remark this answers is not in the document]
jamesmunns: Ah, I forgot to fix this, I mostly needed to loop up how to appropriately debug assert this. Will look into this before merging.

cordyceps/src/sorted_list.rs
jamesmunns: This is "james is paranoid about violating provenance", miri was clean, so I can remove it, but the intent was that we wouldn't start mutating the cursor while a ref to it's value was still live.

cordyceps/src/sorted_list.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        // On every iteration of the loop, we know that the new element should
+        // be placed AFTER the current value of `cursor`, meaning that we need
+        // to decide whether we:
+        //
+        // * just append (if next is null)
+        // * insert between cursor and cursor.next (if elem is < c.next)
+        // * just iterate (if elem >= c.next)
+        loop {
+            // SAFETY: We have exclusive access to the list, we are allowed to
```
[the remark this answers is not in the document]
jamesmunns: Sure! will make a pass.

cordyceps/src/sorted_list.rs
jamesmunns: Nah we totally can do that!

PR #526, opened 2025-04-22
feat(cordyceps): Initial support for non-CAS targets

jamesmunns:
This PR is a first attempt to allow building of `cordyceps` on non-CAS enabled targets, such as `thumbv6m-none-eabi`/Cortex-M0+/rp2040.

Doing this as a first feature, prior to implementing a non-atomic and `critical_section` based version of TransferStack, in service of https://github.com/embassy-rs/embassy/pull/4035.

Rather than making a default-on "CAS" feature, we instead gate on `cfg(target_has_atomic = "ptr")`, which was [stabilized in 1.60](https://github.com/rust-lang/rust/pull/93824), which currently has the meaning of "this target has CAS". This automatically gates-out `TransferStack` and `MpscQueue`, while the "mutable" collections can still be used.

Right now the cfg gating is a little scattershot, I wanted to open this to discuss, if you'd like me to clean things up a bit to group the CAS-ful parts, I'm happy to do so.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,12 +252,17 @@ jobs:
     name: no-atomics compat (cordyceps)
     steps:
     - name: install rust toolchain
-      run: |
-        rustup show
-        rustup target add thumbv6m-none-eabi
+      run: rustup show
     - uses: actions/checkout@v4
     - name: run check
-      run: cargo check -p cordyceps --target thumbv6m-none-eabi
+      # we add the target HERE instead of in the 'install rust toolchain' step
+      # because we need the `rust-toolchain.toml` file to be checked out, otherwise
+      # the target is only added for `stable`. We could also add this to the
+      # `targets` item in the toolchain file, but that would be a waste for all
+      # other tests.
+      run: |
+        rustup target add thumbv6m-none-eabi
+        cargo check -p cordyceps --target thumbv6m-none-eabi
 
   ### maitake ###
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,6 +244,21 @@ jobs:
     - name: run Miri tests
       run: just miri cordyceps
 
+  # run no-atomics target tests
+  cordyceps_no_atomics_compat:
+    needs: changed_paths
+    if: needs.changed_paths.outputs.should_skip != 'true' || !fromJSON(needs.changed_paths.outputs.paths_result).cordyceps.should_skip
+    runs-on: ubuntu-latest
+    name: no-atomics compat (cordyceps)
+    steps:
+    - name: install rust toolchain
+      run: |
+        rustup show
+        rustup target add thumbv6m-none-eabi
+    - uses: actions/checkout@v4
+    - name: run check
+      run: cargo check -p cordyceps --target thumbv6m-none-eabi
+
   ### maitake ###
 
   # test with `--no-default-features`
@@ -313,6 +328,7 @@ jobs:
     - docs
     - cordyceps_loom
     - cordyceps_miri
+    - cordyceps_no_atomics_compat
     - maitake_no_default_features
     - maitake_loom
     - maitake_miri
--- a/cordyceps/src/lib.rs
+++ b/cordyceps/src/lib.rs
@@ -68,18 +68,23 @@ extern crate std;
 pub(crate) mod util;
 
 pub mod list;
+#[cfg(target_has_atomic="ptr")]
 pub mod mpsc_queue;
 pub mod sorted_list;
 pub mod stack;
 
 #[doc(inline)]
 pub use list::List;
+#[cfg(target_has_atomic="ptr")]
 #[doc(inline)]
 pub use mpsc_queue::MpscQueue;
 #[doc(inline)]
 pub use sorted_list::{SortedList, SortedListIter};
 #[doc(inline)]
-pub use stack::{Stack, TransferStack};
+pub use stack::Stack;
+#[cfg(target_has_atomic="ptr")]
+#[doc(inline)]
+pub use stack::TransferStack;
 
 pub(crate) mod loom;
 
--- a/cordyceps/src/lib.rs
+++ b/cordyceps/src/lib.rs
@@ -34,6 +34,9 @@
 //!   [`MpscQueue`]s can be used to efficiently share data from multiple
 //!   concurrent producers with a consumer.
 //!
+//!   This structure is only available if the target supports CAS (Compare and
+//!   Swap) atomics.
+//!
 //! - **[`Stack`]: a mutable, singly-linked first-in, first-out (FIFO)
 //!   stack.**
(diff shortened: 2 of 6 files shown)


hawkw:
Would you mind elaborating a bit on the rationale for doing this by `cfg`-gating out the atomic structures on these targets, rather than using something like `portable-atomic`, which we use elsewhere?

For the record, I'm not suggesting that just disabling these APIs on non-CAS platforms is the wrong approach, I'm just curious about why it's the right one. My guess is that this is because `portable-atomic`'s fallback on these platforms is a mutex-like thing, making the lock free structures suddenly...*not* lock-free, and hand-implementing a critical-section/mutex-y version of the structure would allow better control of where critical sections occur etc?

jamesmunns:
> My guess is that this is because portable-atomic's fallback on these platforms is a mutex-like thing, making the lock free structures suddenly...not lock-free, and hand-implementing a critical-section/mutex-y version of the structure would allow better control of where critical sections occur etc?

Yep! That's exactly it. A naiive "drop in p-a" impl would be decidedly not-great for all structures, so my plan was to make the current lib build, notching out unsupported structures entirely, then in a second pass/PR, re-implement at least `TransferStack`, providing a pretty much separate (but API compatible!) implementation that uses a larger critical section when interacting with the list. This way, we have a single, sliiiightly larger, critical section, rather than a couple of smaller ones.

hawkw:
Yeah, that makes sense. I'm on board with the initial goal of just getting the crate to build for those targets at all!

jamesmunns:
@hawkw is this good to merge, or did you want to take another pass on it?

hawkw:
> @hawkw is this good to merge, or did you want to take another pass on it?

sorry, i wanted to leave a few suggestions on the feature-flagging implementation. give me a sec!

.github/workflows/ci.yml
```diff
@@ -244,6 +244,26 @@ jobs:
     - name: run Miri tests
       run: just miri cordyceps
 
+  # run no-atomics target tests
```
hawkw, line 247: thanks for adding this!

.github/workflows/ci.yml
```diff
@@ -244,6 +244,26 @@ jobs:
...
+    - uses: actions/checkout@v4
+    - name: run check
+      # we add the target HERE instead of in the 'install rust toolchain' step
+      # because we need the `rust-toolchain.toml` file to be checked out, otherwise
+      # the target is only added for `stable`. We could also add this to the
+      # `targets` item in the toolchain file, but that would be a waste for all
+      # other tests.
+      run: |
+        rustup target add thumbv6m-none-eabi
+        cargo check -p cordyceps --target thumbv6m-none-eabi
```
hawkw, line 265: a couple nits: i would prefer to do the `rustup target add` and `cargo check` parts in separate steps, so that it's more obvious which one failed in the github UI. also, rather than leaving the comment discussing how we need to do the target add step separately from the "rustup show" step due to needing the target file, why not just do this:
```suggestion
    - uses: actions/checkout@v4 
    - name: install rust toolchain
      run: rustup target add thumbv6m-none-eabi
    - name: run check
      run: cargo check -p cordyceps --target thumbv6m-none-eabi
```

hawkw: Also, incidentally, i think the current "install rust toolchain" step is wrong _everywhere_, since it's running before checkout and doesn't include the toolchain file. it will always show the default toolchain on the github actions runner, and not the toolchain in our `rust-toolchain.toml`...so it's basically pointless. _And_, furthermore, I think the recent Rustup changes to stop implicitly installing toolchains (rust-lang/rustup#3635) would have broken this if they weren't reverted (rust-lang/rustup#4211), since `rustup show` would no longer install the required toolchain if it's not present. So, we should probably change all the "install rust toolchain" steps by:
- moving them after the checkout step, and
- making them do `rustup toolchain install` rather than `rustup show`.

But, this should be done in a separate PR.

cordyceps/src/lib.rs
```diff
@@ -59,6 +62,15 @@
...
 //!   cleanup.
+//!
+//!   This structure is only available if the target supports CAS (Compare and
+//!   Swap) atomics.
+//!
+//! ## Compatibility
+//!
+//! Rudimentary support for targets without CAS (Compare and Swap) atomics, such as
+//! Cortex-M0+/`thumbv6m-none-eabi`, is provided, however not all structures and
+//! features may be available.
```
hawkw, line 73: Nit, take it or leave it: it would maybe be nice if this explicitly stated that this is detected via `target_has_atomic = "ptr"`, with a reference to the docs for that? It seems like it could be helpful for readers to know exactly what cfg attribute detects this.

cordyceps/src/stack.rs
hawkw, line 1: Take it or leave it: in general, when some type requires a feature or cfg attribute, and also has a bunch of implementations that also are gated by that cfg, I like to stick it in a private submodule and re-export it. That way, only the submodule and the reexport need the cfg attribute, rather than having to put the attribute on every impl block, import, etc used by the cfg-gated type. This makes adding new impl blocks/imports/etc a little less annoying since you don't need to remember to add the cfg everywhere and get caught only when the "no atomics" CI job runs.

Feel free to ignore this advice, but I think this would make life a little easier in the long run.

cordyceps/src/util.rs
hawkw, line 1: similarly, take it or leave it, but it might be a little easier to maintain in the future if we stuck all the "atomics-requiring utils" in a submodule that has the feature gate on it.

in this case i'm less convinced since it's a bunch of mostly-unrelated stuff, but maybe worth considering.

hawkw (commented):
I had a few minor suggestions on how we could make the cfg-gating a bit more ergonomic for future changes to the cfg-gated structures, and a couple small tweaks to the CI job. Other than that, this looks good!

cordyceps/src/lib.rs
hawkw, line 73: oh, one more little nit: i might move this block out of the lib.rs docs, and put it in the [README](https://github.com/hawkw/mycelium/blob/55181917e65b619f756c6d12c26c999d7c8d6646/cordyceps/README.md), which is included at the top of the lib.rs docs here: https://github.com/hawkw/mycelium/blob/55181917e65b619f756c6d12c26c999d7c8d6646/cordyceps/src/lib.rs#L1

the list of individual data structures is in the lib.rs docs rather than in the README that gets included into it because it contains intra-docs links to the different types. but the compatibility section doesn't include any intra-docs links, so it might be nicer to have it in the portion that's also in the README and gets rendered on crates.io etc.

.github/workflows/ci.yml
jamesmunns: Opened #528 to track

cordyceps/src/stack.rs
jamesmunns: Yeah, I'm on board with this, will do.

PR #529, opened 2025-04-27
fix(trace): Fixed infinite line-wrapping

lethalbit:
Within `Writer::write_str` there is a bug in the line wrapping logic, where if the `offset` is ever `0`, it would never force a hard-wrap and cause the poor line wrapping code to spin forever, hopelessly trying its best, but making no progress.

There are 2 cases where `offset` can be `0`, the first is when it can't find any whitespace to wrap on, and the second being when the **very last** character of the wrap boundary was whitespace.

This fixes the primary problem by always forcing an <sub>ugly</sub> hard-wrap at the boundary if `offset` is ever `0`, and updating the `offset` value to be that boundary limit. This then allows the sub-slice of the line to work as expected and not run in circles forever.

### DIFF
--- a/trace/src/lib.rs
+++ b/trace/src/lib.rs
@@ -418,17 +418,27 @@ where
         for line in lines {
             let mut line = line;
             while self.current_line + line.len() >= self.cfg.line_len {
-                let offset = if let Some(last_ws) = line[..self.cfg.line_len - self.current_line]
+                let len_max = self.cfg.line_len - self.current_line;
+                let offset = if let Some(last_ws) = line[..len_max]
                     .chars()
                     .rev()
                     .position(|c| c.is_whitespace())
                 {
                     // found a nice whitespace to break on!
-                    self.writer.write_str(&line[..last_ws])?;
                     last_ws
                 } else {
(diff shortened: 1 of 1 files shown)


hawkw:
Thank you for looking at this! We could probably also write some tests for this stuff (running on the development system with `std`) to catch some of those edge cases, but...I've been too lazy to actually do that. It's totally fine if you are too... :)

lethalbit:
Yeah no problem! I was *ahem* borrowing a big chunk of the code for a thing I'm working on when I found the bug :v 

Some proper tests would be nice, but alas, the spoons and effort needed is not in the budget right now, for me at least.

As for the edge cases I found, there are some very rare cases where line wrapping just goes mental, I suspect there is a subtle off-by-one somewhere, where it imagines that it can to a soft-wrap right after the first character?

This also seemes very inconsistent, as most of the time it would be fine, but then sometimes just, has a conniption

| | |
|-|-|
| image | image |


I've just given up for what i'm doing and force a hard-wrap at the boundary every time, but fixing it would be nice, lol

PR #533, opened 2025-05-20
cordyceps: Add "raw" iter for `List`

### LINKED ISSUE #532: cordyceps: implement handle-based iterator for `List`
If we want to do type-punning shenanigans with `List` like we do for `MpscQueue` in executors, we probably need traversal methods for `List` that give back a `T::Handle` instead of a `Pin<&mut T>`.

CC https://github.com/tweedegolf/cfg-noodle/issues/9#issuecomment-2894145195

This avoids issues with miri if we cast-through a node pointer, because if we do so with a `Pin<&mut T>` we have narrowed the provenance, vs the original pointer.

Open to feedback from @hawkw, otherwise I'll probably plan to implement a version of `Cursor` that yields `T::Handle`s instead. I think we can reuse the same `CoreCursor` underneath the hood, and we don't need mut and immut versions.

jamesmunns:
closes #532

cc https://github.com/tweedegolf/cfg-noodle/issues/9

This is an initial implementation of a "raw" iterator, that allows for type-punning of nodes of a `List`. Unlike `MpscQueue` and `TransferStack`, which give back `T::Handle` elements, the `iter()` and `iter_mut()` methods of `List` instead give `&T` or `Pin<&mut T>`. However, by creating these references, we narrow the provenance of the reference to only the header of a type-punned object, leading to miri stacked borrow violations.

Ideally, we would be able to provide an iterator over something like `&T::Handle`, as we don't want to necessarily give ownership (which `T::Handle` semantically has), but this is difficult to express in an iterator, as we would need to store the `T::Handle` somewhere, and forget it so we don't consume the handle.

Instead, this "RawIter" just gives "raw" `NonNull<T>` pointers to each element, meaning that no intermediary reference is created. This is spicy, but access to the elements still requires `unsafe` to dereference the `NonNull`s, and the iterator itself is still bound to the lifetime of `'list`, similar to `Iter` and `IterMut`.

In https://github.com/jamesmunns/ergot/commit/332085d931635866f88ec188f90b8cd558bb4649, when running `MIRIFLAGS=-Zmiri-disable-isolation cargo +nightly miri test`, miri now passes, whereas in the previous commit (using `iter_mut`) there is a violation that looks like this:

```
test hello ... error: Undefined Behavior: trying to retag from <378652> for SharedReadWrite permission at alloc118585[0x38], but that tag does not exist in the borrow stack for this location
   --> /Users/james/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:428:18
    |
428 |         unsafe { &*self.as_ptr().cast_const() }
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                  |
    |                  trying to retag from <378652> for SharedReadWrite permission at alloc118585[0x38], but that tag does not exist in the borrow stack for this location
    |                  this error occurs as part of retag at alloc118585[0x0..0x60]
    |
    = help: this indicates a potential bug in the program: it performed an invalid operation, but the Stacked Borrows rules it violated are still experimental
    = help: see https://github.com/rust-lang/unsafe-code-guidelines/blob/master/wip/stacked-borrows.md for further information
help: <378652> was created by a SharedReadWrite retag at offsets [0x0..0x38]
   --> /Users/james/personal/ergot/src/lib.rs:91:59
    |
91  | ...   let this: NonNull<SocketHeader> = NonNull::from(unsafe { socket.get_unchecked_mut() });
    |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: BACKTRACE (of the first span) on thread `hello`:
    = note: inside `std::ptr::NonNull::<ergot::socket::endpoint::OwnedSocket<ExampleEndpoint>>::as_ref::<'_>` at /Users/james/.rustup/toolchains/nightly-aarch64-apple-darwin/lib/rustlib/src/rust/library/core/src/ptr/non_null.rs:428:18: 428:46
note: inside `ergot::socket::endpoint::OwnedSocket::<ExampleEndpoint>::send_owned`
   --> /Users/james/personal/ergot/src/socket/endpoint.rs:153:36
    |
153 |         let this: &Self = unsafe { this.as_ref() };
    |                                    ^^^^^^^^^^^^^
note: inside closure
   --> /Users/james/personal/ergot/src/lib.rs:95:25
    |
95  |                         (f)(this, that, &TypeId::of::<T>(), src, dst)
    |                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: inside closure at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mutex-1.0.0/src/lib.rs:80:13: 80:21
    = note: inside closure at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mutex-1.0.0/src/raw_impls.rs:59:27: 59:30
    = note: inside `critical_section::with::<std::option::Option<std::result::Result<(), ()>>, {closure@<mutex::raw_impls::cs::CriticalSectionRawMutex as mutex::ScopedRawMutex>::try_with_lock<std::result::Result<(), ()>, {closure@mutex::BlockingMutex<mutex::raw_impls::cs::CriticalSectionRawMutex, ergot::NetStackInner>::with_lock<std::result::Result<(), ()>, {closure@ergot::NetStack<mutex::raw_impls::cs::CriticalSectionRawMutex>::send_ty<Example>::{closure#0}}>::{closure#0}}>::{closure#0}}>` at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/critical-section-1.2.0/src/lib.rs:248:14: 248:39
    = note: inside `<mutex::raw_impls::cs::CriticalSectionRawMutex as mutex::ScopedRawMutex>::try_with_lock::<std::result::Result<(), ()>, {closure@mutex::BlockingMutex<mutex::raw_impls::cs::CriticalSectionRawMutex, ergot::NetStackInner>::with_lock<std::result::Result<(), ()>, {closure@ergot::NetStack<mutex::raw_impls::cs::CriticalSectionRawMutex>::send_ty<Example>::{closure#0}}>::{closure#0}}>` at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mutex-1.0.0/src/raw_impls.rs:52:13: 62:15
    = note: inside `<mutex::raw_impls::cs::CriticalSectionRawMutex as mutex::ScopedRawMutex>::with_lock::<std::result::Result<(), ()>, {closure@mutex::BlockingMutex<mutex::raw_impls::cs::CriticalSectionRawMutex, ergot::NetStackInner>::with_lock<std::result::Result<(), ()>, {closure@ergot::NetStack<mutex::raw_impls::cs::CriticalSectionRawMutex>::send_ty<Example>::{closure#0}}>::{closure#0}}>` at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mutex-1.0.0/src/raw_impls.rs:70:13: 70:34
    = note: inside `mutex::BlockingMutex::<mutex::raw_impls::cs::CriticalSectionRawMutex, ergot::NetStackInner>::with_lock::<std::result::Result<(), ()>, {closure@ergot::NetStack<mutex::raw_impls::cs::CriticalSectionRawMutex>::send_ty<Example>::{closure#0}}>` at /Users/james/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/mutex-1.0.0/src/lib.rs:76:9: 81:11
note: inside `ergot::NetStack::<mutex::raw_impls::cs::CriticalSectionRawMutex>::send_ty::<Example>`
   --> /Users/james/personal/ergot/src/lib.rs:82:19
    |
82  |           let res = self.inner.with_lock(|inner| {
    |  ___________________^
83  | |             for socket in inner.sockets.iter_mut() {
84  | |                 let port = unsafe { *socket.port.get() };
...   |
105 | |             Err(())
106 | |         });
    | |__________^
note: inside closure
   --> tests/smoke.rs:39:9
    |
39  |         STACK.send_ty::<Example>(src, dst, ExampleEndpoint::REQ_KEY, Example { a: 42, b: 789 }).unwrap();
```

This PR probably needs some more docs and tests, but I wanted to open this up for discussion to see if this is a reasonable thing to want, or if there is another approach we should look into.

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -248,7 +248,7 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
-/// Iterates over the items in a [`List`] by `NonNull<T>` node pointers.
+/// Iterates over the items in a [`List`] as [`NonNull`]`<T>` node pointers.
 ///
 /// Whenever possible, prefer [`Iter`] or [`IterMut`], as they provide
 /// easier-to-hold-right interfaces when iterating over nodes.
@@ -929,7 +929,7 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
     /// Although this method is safe, care must be taken with the items
     /// yielded by the returned iterator.
     ///
-    /// This iterator returns `NonNull<T>` of the individual node elements,
+    /// This iterator returns [`NonNull`]`<T>` of the individual node elements,
     /// rather than references. This is done to allow "type punning" of nodes,
     /// where the creation of a reference could restrict the provenance of the
     /// pointed-to item. This is relevant if your nodes are of a common header
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -248,8 +248,17 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
-/// Iterates over the items in a [`List`] by pointer.
-pub struct RawIter<'list, T: Linked<Links<T>> + ?Sized> {
+/// Iterates over the items in a [`List`] by `NonNull<T>` node pointers.
+///
+/// Whenever possible, prefer [`Iter`] or [`IterMut`], as they provide
+/// easier-to-hold-right interfaces when iterating over nodes.
+///
+/// ## Safety
+///
+/// Although iteration of items is safe, care must be taken with
+/// the returned `NonNull<T>` nodes. See [`List::iter_raw()`] for
+/// more details on safety invariants.
+pub struct IterRaw<'list, T: Linked<Links<T>> + ?Sized> {
     _list: &'list List<T>,
 
     /// The current node when iterating head -> tail.
@@ -914,9 +923,36 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
     }
 
     /// Returns an iterator over the items in this list, by pointer.
+    ///
+    /// ## Safety
+    ///
+    /// Although this method is safe, care must be taken with the items
+    /// yielded by the returned iterator.
+    ///
+    /// This iterator returns `NonNull<T>` of the individual node elements,
+    /// rather than references. This is done to allow "type punning" of nodes,
+    /// where the creation of a reference could restrict the provenance of the
+    /// pointed-to item. This is relevant if your nodes are of a common header
+    /// type, but may have different "body" types you'd like to cast to. This is similar
+    /// to how executors like `maitake` or `tokio` store `Task`s: with a common
+    /// `TaskHeader` but containing varying futured in their bodies. If this is
+    /// NOT a feature you need, consider using [`List::iter()`] or
+    /// [`List::iter_mut()`] instead.
+    ///
+    /// Morally, the elements yielded by this iterator should be treated *as if*
+    /// they were `Pin<NonNull<T>>`, or `Pin<&mut T>`, in that you MUST NOT move
+    /// out of the pointed-to nodes. The nodes are still logically OWNED by the
+    /// list, and must not be removed using this interface. You may still use
+    /// these pointers with whatever provenance they were originally created with,
+    /// allowing for type-punning to a larger type with a common header base.
+    ///
+    /// Unfortunatly we [cannot create] a `Pin<NonNull<T>>`, so you must use the
+    /// pointers yielded by this iterator carefully.
+    ///
+    /// [cannot create]: https://github.com/rust-lang/rust/issues/54815
     #[must_use]
-    pub fn raw_iter(&self) -> RawIter<'_, T> {
-        RawIter {
+    pub fn iter_raw(&mut self) -> IterRaw<'_, T> {
+        IterRaw {
             _list: self,
             curr: self.head,
             curr_back: self.tail,
@@ -1407,7 +1443,7 @@ impl<T: Linked<Links<T>> + ?Sized> iter::FusedIterator for IterMut<'_, T> {}
 
 // === impl RawIter ====
 
-impl<T: Linked<Links<T>> + ?Sized> Iterator for RawIter<'_, T> {
+impl<T: Linked<Links<T>> + ?Sized> Iterator for IterRaw<'_, T> {
     type Item = NonNull<T>;
 
     fn next(&mut self) -> Option<Self::Item> {
@@ -1434,14 +1470,14 @@ impl<T: Linked<Links<T>> + ?Sized> Iterator for RawIter<'_, T> {
     }
 }
 
-impl<T: Linked<Links<T>> + ?Sized> ExactSizeIterator for RawIter<'_, T> {
+impl<T: Linked<Links<T>> + ?Sized> ExactSizeIterator for IterRaw<'_, T> {
     #[inline]
     fn len(&self) -> usize {
         self.len
     }
 }
 
-impl<T: Linked<Links<T>> + ?Sized> DoubleEndedIterator for RawIter<'_, T> {
+impl<T: Linked<Links<T>> + ?Sized> DoubleEndedIterator for IterRaw<'_, T> {
     fn next_back(&mut self) -> Option<Self::Item> {
         if self.len == 0 {
             return None;
@@ -1460,7 +1496,7 @@ impl<T: Linked<Links<T>> + ?Sized> DoubleEndedIterator for RawIter<'_, T> {
     }
 }
 
-impl<T: Linked<Links<T>> + ?Sized> iter::FusedIterator for RawIter<'_, T> {}
+impl<T: Linked<Links<T>> + ?Sized> iter::FusedIterator for IterRaw<'_, T> {}
 
 
 // === impl IntoIter ===
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -950,6 +950,16 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
     /// pointers yielded by this iterator carefully.
     ///
     /// [cannot create]: https://github.com/rust-lang/rust/issues/54815
+    ///
+    /// ## Example
+    ///
+    /// For an end-to-end example of the kind of "spooky type punning" this interface
+    /// aims to allow, check out the [test for `iter_raw`], which demonstrates the
+    /// ability to call a dynamic "print" function on any item in the list, where
+    /// all items are of differing types (but all implement the [`Debug`][core::fmt::Debug]
+    /// trait).
+    ///
+    /// [test for `iter_raw`]: https://github.com/hawkw/mycelium/blob/main/cordyceps/src/list/tests/iter_raw.rs
     #[must_use]
     pub fn iter_raw(&mut self) -> IterRaw<'_, T> {
         IterRaw {
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -248,6 +248,22 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
+/// Iterates over the items in a [`List`] by pointer.
+pub struct RawIter<'list, T: Linked<Links<T>> + ?Sized> {
+    _list: &'list List<T>,
+
+    /// The current node when iterating head -> tail.
+    curr: Link<T>,
+
+    /// The current node when iterating tail -> head.
+    ///
+    /// This is used by the [`DoubleEndedIterator`] impl.
+    curr_back: Link<T>,
+
+    /// The number of remaining entries in the iterator.
+    len: usize,
+}
+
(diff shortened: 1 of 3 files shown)


jamesmunns:
Open questions/thoughts:

* Do we need `raw_iter()` to also be marked `unsafe`? As I mention above, interacting with the nodes still requires unsafe, so we could maybe avoid it, but it might be good to note that morally users should treat the ptrs as if they were `Pin<&mut T>`.
* Do we need a `raw_iter_mut()`, or raw versions of `Cursor` or `CursorMut`? I could see wanting to operate on the list itself while working with raw ptrs. This could be a future PR, but in case we want the naming to be consistent, we might want to give it a think
* Is there a better name than `RawIter`?
* Is there a better type to yield than `NonNull<T>`?

cordyceps/src/list.rs
```diff
@@ -1378,6 +1405,64 @@ impl<T: Linked<Links<T>> + ?Sized> DoubleEndedIterator for IterMut<'_, T> {
 
 impl<T: Linked<Links<T>> + ?Sized> iter::FusedIterator for IterMut<'_, T> {}
 
+// === impl RawIter ====
+
+impl<T: Linked<Links<T>> + ?Sized> Iterator for RawIter<'_, T> {
+    type Item = NonNull<T>;
```
hawkw, line 1411: Hmm, so, in the PR description, you say that this iterates over the `T::Handle` type of the `Linked` implementation for `T`, but here, we actually always return `NonNull<T>`s.

For the record, I think this is more correct. The intent is that `T::Handle` is an _owning_ reference; if `T` isn't something like a static, dropping the `Handle` type should deallocate the `T`, so we would expect it to be implemented for types like `Pin<Box<T>>` and similar. It would definitely be wrong for the raw iterator to yield `Box<T>`s containing elements that are still in the list, as the iterator items would drop them. So, I think the use of `NonNull` here is correct.

cordyceps/src/list.rs
```diff
@@ -897,6 +913,17 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         }
     }
 
+    /// Returns an iterator over the items in this list, by pointer.
+    #[must_use]
+    pub fn raw_iter(&self) -> RawIter<'_, T> {
```
hawkw, line 918: hmm, should this mutably borrow `self`? since the entries can be mutated through the iterator..

it occurs to me that we _could_ even have a `RawIter` that immutably borrows the list and hands out `*const T`s, and a `RawIterMut` that mutably borrows the list and hands out `*mut T`s or `NonNull<T>`s. but, the `*const T` means we lose out on the `NonNull` guarantee and niche optimization...i am once again wishing that libcore had a `ConstNonNull` type thing...

hawkw, line 918: also, turbo nit: i might consider calling this `iter_raw` rather than `raw_iter`, which feels more consistent with e.g. `iter_mut`

cordyceps/src/list.rs
```diff
@@ -897,6 +913,17 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
         }
     }
 
+    /// Returns an iterator over the items in this list, by pointer.
```
hawkw, line 916: i don't think that the act of _constructing_ a `RawIter` should be unsafe --- it's analogous to constructing a raw pointer being safe, but dereferencing it being unsafe. however, I _do_ think the docs for this method should discuss the safety invariants for using the raw pointers handed out by the `RawIter`. in particular, you definitely cannot move out of them --- they are "logically pinned" even though they cannot be `Pin<NonNull<T>>`[^1] --- and we ought to state that here. 

[^1]: see https://github.com/rust-lang/rust/issues/54815

cordyceps/src/list.rs
hawkw: i think this _has_ to mutably borrow `self` since otherwise, it allows constructing a mutable reference from an immutable reference --- does miri not get mad if you do that? we might consider adding the `*const T` iterator as well.

cordyceps/src/list.rs
```diff
@@ -248,6 +248,22 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
+/// Iterates over the items in a [`List`] by pointer.
+pub struct RawIter<'list, T: Linked<Links<T>> + ?Sized> {
```
hawkw, line 252: part of me kind of wonders if we should just go ahead and make a full "RawCursor" type instead of an iterator. but i'm not actually sure if that's necessary --- i dunno if you would necessarily need to do all the other cursor operations using a raw pointer-y thing.

hawkw (commented):
i definitely agree with the rationale behind this. i had some thoughts on the interface.

cordyceps/src/list.rs
jamesmunns: I think in my case it doesn't complain, likely due to some unsafe cell indirection, but it makes sense if the list owns all nodes it should require mut to allow full raw access.

cordyceps/src/list.rs
jamesmunns: I think the issue pointed at handle, in the pr I mention that nonnull is probably the practical right choice.

cordyceps/src/list.rs
jamesmunns: I'm cool with either, I went back and forth.

cordyceps/src/list.rs
jamesmunns: I think I will want it, some sockets are one shot, so being able to use the ptr, then remove the node from the list would be useful. I might do that in a follow up PR tho.

jamesmunns:
@hawkw addressed all review comments, and added a basic test that is a fun mini-example of type punning :)

cordyceps/src/list.rs
hawkw: Oh, I read the wrong thing, whoops.

cordyceps/src/list.rs
```diff
@@ -897,6 +922,44 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
...
     }
 
+    /// Returns an iterator over the items in this list, by pointer.
+    ///
+    /// ## Safety
+    ///
+    /// Although this method is safe, care must be taken with the items
+    /// yielded by the returned iterator.
+    ///
+    /// This iterator returns `NonNull<T>` of the individual node elements,
```
hawkw, line 932: teensy nit:
```suggestion
    /// This iterator returns [`NonNull`]`<T>` of the individual node elements,
```

cordyceps/src/list/tests/iter_raw.rs
```diff
@@ -0,0 +1,131 @@
...
+    // We use iter_raw instead of iter_mut to allow type punning
+    for node in list.iter_raw() {
+        // obtain the vtable
+        let vtable: &'static VTable = unsafe { node.as_ref().vtable };
+        // type erase, call vtable method
+        let node: NonNull<()> = node.cast();
+        (vtable.print)(node);
+    }
+
+    // Try running this with `cargo +nightly miri test -- iter_raw --nocapture`, it's pretty neat :)
```
hawkw, line 128: i wonder if we ought to also copy some of this test into an example? the docs coudl also reference it...

cordyceps/src/list.rs
```diff
@@ -897,6 +922,44 @@ impl<T: Linked<Links<T>> + ?Sized> List<T> {
...
+    /// rather than references. This is done to allow "type punning" of nodes,
+    /// where the creation of a reference could restrict the provenance of the
+    /// pointed-to item. This is relevant if your nodes are of a common header
+    /// type, but may have different "body" types you'd like to cast to. This is similar
+    /// to how executors like `maitake` or `tokio` store `Task`s: with a common
+    /// `TaskHeader` but containing varying futured in their bodies. If this is
```
hawkw, line 938: maybe worth linking to one of those examples? although maybe it's sprayed too far across those codebases to be easily referenceable. we could also try to turn some of your nice test into an example we could reference here.

cordyceps/src/list.rs
```diff
@@ -248,6 +248,31 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
     len: usize,
 }
 
+/// Iterates over the items in a [`List`] by `NonNull<T>` node pointers.
```
hawkw, line 251: nit:
```suggestion
/// Iterates over the items in a [`List`] as [`NonNull`]`<T>` node pointers.
```

hawkw (approved):
Looks good to me --- left a couple last suggestions.

cordyceps/src/list.rs
jamesmunns: We could, though those use different structures (mpscqueue, transferstack), so it might not be the cognitive load to follow.

the test could be useful, it requires a lot of moving bits as a minimal example tho.

PR #536, opened 2025-06-02
Correct borrow, and impl Send for RawIter

### LINKED ISSUE #535: cordyceps: Should `list::IterRaw` impl Send?
Reported in https://github.com/tweedegolf/cfg-noodle/issues/14#issuecomment-2901241711, holding the iterator across an await point (e.g. using an async mutex) means that the RawIter might need to impl send. IMO this is probably fine since we are bound by the lifetime of the mut borrow of the list.

jamesmunns:
Fixes #535

This change fixes the two things noted in the #535: RawIter now holds an `&mut` to the List, and also now implements Send.

I think we could also implement Send for Iter/IterMut, but I'm less confident about that.

I also am not certain whether the unsafe `Send` impl should be gated on `T: Send`, I opted not to because we never take ownership of a node (e.g. a `pop`) using the RawIter interface, so we're only ever "visiting" a pinned node, so requiring `T: Send` didn't seem necessary. I'm open to more confident logic around that.

I also added a util helper to assert that RawIter remains Send.

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -259,7 +259,7 @@ pub struct Iter<'list, T: Linked<Links<T>> + ?Sized> {
 /// the returned `NonNull<T>` nodes. See [`List::iter_raw()`] for
 /// more details on safety invariants.
 pub struct IterRaw<'list, T: Linked<Links<T>> + ?Sized> {
-    _list: &'list List<T>,
+    _list: &'list mut List<T>,
 
     /// The current node when iterating head -> tail.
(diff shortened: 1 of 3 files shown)


jamesmunns:
(I still think we should merge this, in case some users don't need to hold the yielded ptrs across await point, and to fix the &mut soundness hole).

hawkw (approved):
okay yeah we should probably just merge this, thanks and sorry i forgot about it for, uh, 7 months... 😅

Issue #528, opened 2025-04-25
chore: Fix CI install rust toolchain


jamesmunns:
Also, incidentally, i think the current "install rust toolchain" step is wrong _everywhere_, since it's running before checkout and doesn't include the toolchain file. it will always show the default toolchain on the github actions runner, and not the toolchain in our `rust-toolchain.toml`...so it's basically pointless. _And_, furthermore, I think the recent Rustup changes to stop implicitly installing toolchains (rust-lang/rustup#3635) would have broken this if they weren't reverted (rust-lang/rustup#4211), since `rustup show` would no longer install the required toolchain if it's not present. So, we should probably change all the "install rust toolchain" steps by:
- moving them after the checkout step, and
- making them do `rustup toolchain install` rather than `rustup show`.

But, this should be done in a separate PR.

_Originally posted by @hawkw in https://github.com/hawkw/mycelium/pull/526#discussion_r2060638703_

PR #538, opened 2025-07-01
fix(maitake-sync): no-cas maitake-sync

jamesmunns:
There is somewhat confusingly both `loom::sync::atomic` and `loom::atomic`, only the latter of which uses `portable-atomic`. This leads to maitake-sync not building on no-std crates.

This PR fixes this, and adds a test to ensure we do not regress maitake-sync in this regard.

~We may want to do further cleanup of how we do loom exports, but this is fine (if not still confusing) for now.~

I changed this to just stick `loom::atomic` inside `loom::sync`, so now `loom::sync::atomic` does what I think the original intent was.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -264,6 +264,26 @@ jobs:
         rustup target add thumbv6m-none-eabi
         cargo check -p cordyceps --target thumbv6m-none-eabi
 
+  # run no-atomics target tests
+  maitake_sync_no_atomics_compat:
+    needs: changed_paths
+    if: needs.changed_paths.outputs.should_skip != 'true' || !fromJSON(needs.changed_paths.outputs.paths_result).maitake.should_skip
+    runs-on: ubuntu-latest
+    name: no-atomics compat (maitake-sync)
+    steps:
+    - name: install rust toolchain
+      run: rustup show
+    - uses: actions/checkout@v4
+    - name: run check
+      # we add the target HERE instead of in the 'install rust toolchain' step
+      # because we need the `rust-toolchain.toml` file to be checked out, otherwise
+      # the target is only added for `stable`. We could also add this to the
+      # `targets` item in the toolchain file, but that would be a waste for all
+      # other tests.
+      run: |
+        rustup target add thumbv6m-none-eabi
+        cargo check -p maitake-sync --target thumbv6m-none-eabi --no-default-features --features=critical-section
+
   ### maitake ###
 
   # test with `--no-default-features`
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -264,6 +264,26 @@ jobs:
...
+    - uses: actions/checkout@v4
+    - name: run check
+      # we add the target HERE instead of in the 'install rust toolchain' step
+      # because we need the `rust-toolchain.toml` file to be checked out, otherwise
+      # the target is only added for `stable`. We could also add this to the
+      # `targets` item in the toolchain file, but that would be a waste for all
+      # other tests.
+      run: |
+        rustup target add thumbv6m-none-eabi
+        cargo check -p maitake-sync --target thumbv6m-none-eabi --no-default-features --features=critical-section
```
hawkw, line 285: hmm, didn't we previously determine that [the `install rust toolchain` step is Just Wrong Everywhere](https://github.com/hawkw/mycelium/issues/528) since it runs before the toolchain file is checked out? we should fix that everywhere...but in the meantime, I think we should do the following here:

- change this so that there's a separate toolchain step that just installs `thumbv6m-none-eabi` after checkout
- edit the comment to just say "we don't install the toolchain in the toolchain file because we don't need it for every build"

```suggestion
    steps:
    - uses: actions/checkout@v4
    - name: install rust toolchain
      # We add the target here, rather than in the `targets` item in the 
      # toolchain file, because it would be a waste for all other CI jobs to
      # have to install it.
      run: rustup target add thumbv6m-none-eabi
    - name: run check
      run: cargo check -p maitake-sync --target thumbv6m-none-eabi --no-default-features --features=critical-section
```

.github/workflows/ci.yml
```diff
@@ -264,6 +264,26 @@ jobs:
...
+
   ### maitake ###
```
hawkw, line 287: can we move this line to above the newly-added CI job? i wanted to include a little section header for each crate that gets tested, and now the new Maitake check isn't under that heading :P

.github/workflows/ci.yml
```diff
@@ -264,6 +264,26 @@ jobs:
         rustup target add thumbv6m-none-eabi
         cargo check -p cordyceps --target thumbv6m-none-eabi
 
+  # run no-atomics target tests
+  maitake_sync_no_atomics_compat:
```
hawkw, line 268: If we want this job to be required for PRs to merge, we need to also add it to [the `needs:` list for the `all_systems_go` job](https://github.com/hawkw/mycelium/blob/fb9bd10d1e5dcf4c2273ff8f0f679f53d6044820/.github/workflows/ci.yml#L345-L359), which is what actually gates merging a branch. Otherwise, PRs will be able to automerge even if we've broken this check, which seems un-desirable.

hawkw (changes requested):
There are a few issues with the CI config I'd like to fix before merging this. Otherwise, this looks good to me!

hawkw (approved):
looks good to me pending CI!

PR #541, opened 2025-07-14
chore(maitake-sync): restore missing commit

jamesmunns:
When maitake-sync v0.2.1 was released, it looks like https://github.com/hawkw/mycelium/commit/339fd745140513f802b67ec4ff9fe25987d8fae5 was never pushed to main, though the tag was.

This PR merges back the one commit, as an alternative to doing a hard rebase.


hawkw (approved):
my bad, thanks for fixing this

PR #543, opened 2025-08-03
fix stuff that should have been in #542

hawkw:
somehow PR #542 automerged two weeks ago, despite tests failing. our CI situation is messed up. i gotta fix that. anyway here's wonderwall

### DIFF
--- a/hal-x86_64/src/segment.rs
+++ b/hal-x86_64/src/segment.rs
@@ -524,13 +524,32 @@ impl SystemDescriptor {
     const BASE_HIGH: bits::Pack64 = bits::Pack64::least_significant(32);
     const BASE_HIGH_PAIR: Pair64 = Self::BASE_HIGH.pair_with(base::HIGH);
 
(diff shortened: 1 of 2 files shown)


hawkw:
oh, i see why it automerged: currently we have one CI job that does nothing but depends on all the test jobs, called `all_systems_go`. we use that to gate whether a PR can merge, so that the list of required tests is defined in the YAML rather than in the settings UI: https://github.com/hawkw/mycelium/blob/c045b6e7bf1bcf6e83720f0c926f20c06c496aab/.github/workflows/ci.yml#L340-L359

*however*, it seems like the behavior of the GitHub Actions `needs:` thingy is to _skip_ the job if something it depends on fails, rather than failing it:
image
and, i guess, if a job is skipped, it doesn't actually count against "Required Checks" when determining whether a PR can merge. so i guess this means our CI just totally doesn't stop broken PRs from merging at all. that's so cool lmao.

PR #546, opened 2025-08-16
feat(x86_64): dump control registers in oopses and the kernel shell

hawkw:
Faults:
image
...and panics: 
image

Kernel shell dump commands:
image


hawkw:
well, that sucks: https://github.com/hawkw/mycelium/actions/runs/17011279332/job/48227645173?pr=546#step:3:12

PR #548, opened 2025-08-23
16550 input plus UART shells

iximeow:
this is messy but i whacked it with a hammer til it at least does the rightish things. now if you `qemu -display none -serial stdio` you can still get an interactive shell to blow up mycelium with.

image

### DIFF
--- a/cordyceps/src/list.rs
+++ b/cordyceps/src/list.rs
@@ -70,7 +70,7 @@ pub use self::cursor::{Cursor, CursorMut};
 ///
 ///     /// Convert an owned `Handle` into a raw pointer
 ///     fn into_ptr(handle: Pin<Box<Entry>>) -> NonNull<Entry> {
-///        unsafe { NonNull::from(Box::leak(Pin::into_inner_unchecked(handle))) }
+///        unsafe { NonNull::new_unchecked(Box::into_raw(Pin::into_inner_unchecked(handle))) }
 ///     }
 ///
 ///     /// Convert a raw pointer back into an owned `Handle`.
@@ -126,7 +126,7 @@ pub use self::cursor::{Cursor, CursorMut};
 /// # unsafe impl Linked<list::Links<Entry>> for Entry {
 /// #     type Handle = Pin<Box<Self>>;
 /// #     fn into_ptr(handle: Pin<Box<Entry>>) -> NonNull<Entry> {
-/// #        unsafe { NonNull::from(Box::leak(Pin::into_inner_unchecked(handle))) }
+/// #        unsafe { NonNull::new_unchecked(Box::into_raw(Pin::into_inner_unchecked(handle))) }
 /// #     }
 /// #     unsafe fn from_ptr(ptr: NonNull<Entry>) -> Pin<Box<Entry>> {
 /// #         Pin::new_unchecked(Box::from_raw(ptr.as_ptr()))
@@ -180,7 +180,7 @@ pub use self::cursor::{Cursor, CursorMut};
 /// # unsafe impl Linked<list::Links<Entry>> for Entry {
 /// #     type Handle = Pin<Box<Self>>;
 /// #     fn into_ptr(handle: Pin<Box<Entry>>) -> NonNull<Entry> {
-/// #        unsafe { NonNull::from(Box::leak(Pin::into_inner_unchecked(handle))) }
+/// #        unsafe { NonNull::new_unchecked(Box::into_raw(Pin::into_inner_unchecked(handle))) }
 /// #     }
 /// #     unsafe fn from_ptr(ptr: NonNull<Entry>) -> Pin<Box<Entry>> {
 /// #         Pin::new_unchecked(Box::from_raw(ptr.as_ptr()))
--- a/maitake-sync/src/wait_map.rs
+++ b/maitake-sync/src/wait_map.rs
@@ -614,6 +614,220 @@ impl<K: PartialEq, V, Lock: ScopedRawMutex> WaitMap<K, V, Lock> {
         }
     }
 
+    /// Asynchronously poll the given function `f` until a condition occurs,
+    /// using the [`WaitMap`] to only re-poll when notified.
+    ///
+    /// This can be used to implement a "wait loop", turning a "try" function
+    /// (e.g. "try_recv" or "try_send") into an asynchronous function (e.g.
+    /// "recv" or "send").
+    ///
+    /// In particular, this function correctly *registers* interest in the [`WaitMap`]
+    /// prior to polling the function, ensuring that there is not a chance of a race
+    /// where the condition occurs AFTER checking but BEFORE registering interest
+    /// in the [`WaitMap`], which could lead to deadlock.
+    ///
+    /// This is intended to have similar behavior to `Condvar` in the standard library,
+    /// but asynchronous, and not requiring operating system intervention (or existence).
+    ///
+    /// In particular, this can be used in cases where interrupts or events are used
+    /// to signify readiness or completion of some task, such as the completion of a
+    /// DMA transfer, or reception of an ethernet frame. In cases like this, the interrupt
+    /// can wake the queue, allowing the polling function to check status fields for
+    /// partial progress or completion, and also return the status flags at the same time.
+    ///
+    /// Consider using [`Self::wait_for()`] if your function does not return a value.
+    ///
+    /// Consider using [`WaitMap::wait_for_value()`](super::wait_map::WaitMap::wait_for_value)
+    /// if you do not need multiple waiters.
+    ///
+    /// * [`Ok`]`(T)` if the closure returns [`Some`]`(T)`.
+    /// * [`Err`]`(`[`Closed`](crate::Closed)`)` if the [`WaitMap`] is closed.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// # use tokio::task;
+    /// # #[tokio::main(flavor = "current_thread")]
+    /// # async fn test() {
+    /// use std::sync::Arc;
+    /// use maitake_sync::WaitMap;
+    /// use std::sync::atomic::{AtomicBool, Ordering};
+    ///
+    /// let map = Arc::new(WaitMap::new());
+    /// let wake1 = Arc::new(AtomicBool::new(false));
+    /// let wake2 = Arc::new(AtomicBool::new(false));
+    ///
+    /// let waiter1 = task::spawn({
+    ///     // clone items to move into the spawned task
+    ///     let map = map.clone();
+    ///     let wake = wake1.clone();
+    ///     async move {
+    ///         map.wait_for(1, || wake.load(Ordering::Relaxed)).await;
+    ///         println!("received wakeup 1!");
+    ///     }
+    /// });
+    ///
+    /// let waiter2 = task::spawn({
+    ///     // clone items to move into the spawned task
+    ///     let map = map.clone();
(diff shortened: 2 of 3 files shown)


hal-x86_64/src/interrupt.rs
```diff
@@ -727,6 +729,63 @@ mod isr {
...
+    pub(super) extern "x86-interrupt" fn com1<H: Handlers<Registers>>(_regs: Registers) {
+        use crate::serial::Pc16550dInterrupt;
+        let port = crate::serial::com1().expect("can port??");
+        // This block must absolutely not log to COM1, because logging will try to lock COM1 too
+        // and then deadlock.
+
+        // TODO(ixi): If we were interrupted while writing to the serial port, someone else may
+        // have already locked the port, at which point this will deadlock. That's bad. I think we
+        // need to split this into read-lock and write-lock, where someone can hold the write lock
+        // concurrent with us grabbing the read lock here.
```
iximeow, line 741: lol.

hal-x86_64/src/interrupt.rs
```diff
@@ -727,6 +729,50 @@ mod isr {
...
 
+    pub(super) extern "x86-interrupt" fn com1<H: Handlers<Registers>>(_regs: Registers) {
+        let port = crate::serial::com1().expect("can port??");
+        // This block must absolutely not log to COM1, because logging will try to lock COM1 too
+        // and then deadlock.
+        let (b, isr_code, res) = {
+            // TODO: this should be over in the 16550 driver.
+            let mut port = port.lock().set_non_blocking();
+            let isr = port.get_isr();
+            let isr_code = isr & 0b1111;
```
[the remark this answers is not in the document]
iximeow: yeah this (and `get_isr`) turned into the more reasonable [check_interrupt_type](https://github.com/hawkw/mycelium/pull/548/commits/73ec3aa530c0155a2150fb1763f8b2f152b53ce3) once i found an actual source of "how it works" i'm willing to cite ([thank you wiki.qemu.org](https://github.com/hawkw/mycelium/pull/548/commits/73ec3aa530c0155a2150fb1763f8b2f152b53ce3))

hal-x86_64/src/interrupt.rs
hawkw: oh no, oh fuck

src/lib.rs
```diff
@@ -220,8 +215,50 @@ async fn keyboard_demo() {
...
+            other => tracing::warn!(?other, "you typed something weird"),
+        }
+        if newline {
+            return line;
         }
     }
 }
+
+type ReadLineFn = fn(alloc::string::String) -> core::pin::Pin<
+    alloc::boxed::Box<dyn core::future::Future<Output = alloc::string::String> + Send>
```
hawkw, line 247: boxing a future for every line makes me a bit sad since i know my allocator isn't very good...

src/drivers/serial_input.rs
```diff
@@ -0,0 +1,47 @@
...
+pub static SERIAL_INPUTS: InitOnce<alloc::vec::Vec<SerialInput>> = InitOnce::uninitialized();
+
+impl SerialInput {
+    pub async fn next_byte(&self) -> u8 {
+        if let Some(byte) = self.buf.pop() {
+            return byte.expect("buffer contains somes");
+        }
+
+        self.waiters.wait().await.expect("serial waiters should never be closed FOR NOW");
+        self.buf.pop().expect("just got woken up wtf").expect("buffer contains somes")
```
hawkw, line 29: oh my

src/drivers/serial_input.rs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    pub fn handle_input(&self, byte: u8) {
+        if let Err(byte) = self.buf.push(Some(byte)) {
+            // TODO(ixi): We were trying to handle serial console input but we're so backed up that
+            // we don't have space in the buffer to accept a byte. We should probably mask the
+            // interrupt until space in the buffer is available, but instead for now we'll eat the
+            // byte and drop it.
```
hawkw, line 37: cool great sure why not

hal-x86_64/src/serial.rs
```diff
@@ -248,6 +278,39 @@ impl<'a> Lock<'a> {
 }
 
 impl<B> Lock<'_, B> {
+    pub fn check_interrupt_type(&mut self) -> io::Result<Option<Pc16550dInterrupt>> {
+        // IIR bits 0 through 3 describe what happened to produce an interrupt, with bits 4 and 5
+        // always 0, and bits 6 and 7 set to 1 if `FCR0=1`.
```
hawkw, line 283: what does FCR0 mean? i assume it's for Flow ContR0l...?

hal-x86_64/src/serial.rs
```diff
@@ -248,6 +278,39 @@ impl<'a> Lock<'a> {
...
+        // TODO(ixi): probably should check IIR bits 6 and 7? punting on everything related to FIFO
+        // though.
+
+        let interrupt = match self.inner.inner.iir() & 0b1111 {
+            0b0000 => Pc16550dInterrupt::ModemStatus,
+            0b0001 => { return Ok(None); }
+            0b0010 => Pc16550dInterrupt::TransmitterHoldingRegEmpty,
+            0b0100 => Pc16550dInterrupt::ReceivedDataAvailable,
+            0b0110 => Pc16550dInterrupt::ReceiverLineStatus,
+            0b1100 => Pc16550dInterrupt::CharacterTimeout,
```
hawkw, line 302: a thought: you could write this as `bits if bits == Pc16550dInterrupt::WhicheverVariant as u8` so that the bit patterns are unfuckupable...might compile worse though idk. hopefully this is a nop if it isn't == 0b0001...

src/lib.rs
iximeow: oh god i typed Box and still forgot that that means an alloc anyway for each line. damnit

anyway probably? i'll take another shot at making it more generator-y where you give it a borrow and it eventually yields `Ok(())` when the line is done or something.

hal-x86_64/src/serial.rs
iximeow: yeah... one of the enum crates gives you a nice to/from primitive function too right? i should just do that

hal-x86_64/src/interrupt.rs
hawkw: hey, couldn't we also just mask out the RX interrupt while we're in a write and unmask it when we drop the lock?

iximeow: alright so: splitting the port into read/write parts wasn't all that bad once i sat down to do it, and i decided to do that because i couldn't convince myself that we'll _never_ have some lock fuckiness if we have the one lock and IRQ masking/unmasking. mostly because we can't atomically get IRQ masking state and change it.

if we do:
```
lock()
mask()
// op
unmask()
unlock()
```
then the period between unmask/unlock means an interrupt can fire, lock, and deadlock. if we mask/lock/op/unlock/unmask then another thread might unmask interrupts right after we masked them but before we locked, and then an interrupt could fire and deadlock. wherever the masking happens, it's not great.

another option is that the ISR should just never lock the port, but that means we need to be able to take the interrupt and plumb it to something we promptly switch to that can implement the read. that'll _work_ but is an even more absurd amount of plumbing to read a byte so it feels super bad.

having read and write halves of the lock means that as long as interrupt only reads, we can all at once have:

* no deadlock if we log (or panic!) in the COM1 ISR (because it doesn't block writing)
* no need to bypass locks in ISRs
* immediate reads when bytes are ready
* a clearish procedure to get exclusive ownership of a port: lock writing (block other processors), mask interrupts, lock reading (block self until outstanding interrupts are done), and now you own the port with interrupts off. this in reverse is a safe (non-deadlocky) return to sharing the port.

i didn't implement the last bit because i don't have a reason to, but i did jot this down [in more detail](https://github.com/hawkw/mycelium/pull/548/commits/7b5de366530c9d09e0d45197b6b8597488ee6a58#diff-c3859f73039c5f99531dc8a5f5c07a72d9d21718be5c8ac4314ca5c0cc9257c0R10-R49) in case someone comes back to it

hopefully this is legible and not 2am wordspew and if it is wordspew sorry about that

hawkw: I think having separate read- and write-side locks for the port seems like the right approach for the UART.

hawkw (approved):
okay, i think i like the split locks design. there's some stuff that may need to change in the longer term (especially if we want to make the driver more async-friendly) but i think this looks good for now. are you gonna wanna merge this PR or are you still tidying up?

hal-x86_64/src/interrupt.rs
```diff
@@ -727,6 +729,62 @@ mod isr {
...
+                }
+            }
+            Some(other) => {
+                // We're gonna warn about the unhandled 16550 interrupt, but unlock the port for
+                // reading so we don't hold up anyone else waiting to read-lock for management
+                // operations.
+                core::mem::drop(port);
+
+                tracing::warn!("Unhandled 16550 interrupt cause: {other:?}");
+            }
```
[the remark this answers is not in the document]
iximeow: yeah that's probably smart. i'd initially had this `panic!()` and just kinda partially unwound that. an unhandled interrupt cause is Not Great and A Bug but it probably shouldn't immediately kill the computer

hal-x86_64/src/serial.rs
```diff
@@ -31,25 +80,67 @@ pub fn com4() -> Option<&'static Port> {
...
-    inner: LockInner<'a>,
+    inner: WriteLockInner<'a>,
     _is_blocking: PhantomData<B>,
 }
 
-struct LockInner<'a> {
-    inner: MutexGuard<'a, Registers, Spinlock>,
+struct ReadLockInner<'a> {
+    inner: MutexGuard<'a, ReadRegisters, Spinlock>,
+}
```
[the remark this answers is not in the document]
iximeow: i figured the symmetry was nice but i can't really imagine a reason for the read lock needing a drop impl so the newtype's a gonetype now

hal-x86_64/src/serial.rs
```diff
@@ -31,25 +80,67 @@ pub fn com4() -> Option<&'static Port> {
...
+
 // #[derive(Debug)]
 pub struct Port {
-    inner: Mutex<Registers, Spinlock>,
+    read_inner: Mutex<ReadRegisters, Spinlock>,
+    write_inner: Mutex<WriteRegisters, Spinlock>,
 }
 
+/// A lock for the read parts of a serial port. The read and write parts of this port state are
+/// described in the documentation for this module.
```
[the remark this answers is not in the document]
iximeow: as a chronic bad rustdoc user i honestly just didn't know how to spell the link. `[...](crate::serial#implementation)` seems to render right though? so ty for the nudge

hal-x86_64/src/serial.rs
iximeow: add a bit more in the module docs because i forgot where this stuff even is defined or how the datasheet talks about bits (and it doesn't... really clarify itself either), but "FIFO Control Register bit 0" which is a very 1995 way to say it

iximeow:
so, i admit defeat on the async "read line" thingy not involving an alloc, i just can't figure out how to spell it in a way Rust is happy with. if you're ok with that, lets shipit?

hawkw:
> so, i admit defeat on the async "read line" thingy not involving an alloc, i just can't figure out how to spell it in a way Rust is happy with. if you're ok with that, lets shipit?

sure, whatever. i might try and mess with it later. thanks!

Issue #550, opened 2025-09-15
`Writer::write_str` does an infinite loop if a line is longer than line_len


iximeow:
```rust
            while self.current_line + line.len() >= self.cfg.line_len {
                let offset = if let Some(last_ws) = line[..self.cfg.line_len - self.current_line]
                    .chars()
                    .rev()
                    .position(|c| c.is_whitespace())
                {
                    // found a nice whitespace to break on!
                    self.writer.write_str(&line[..last_ws])?;
                    last_ws
                } else {
                    0
                };
                self.writer.write_char('\n')?;
                self.write_newline()?;
                self.writer.write_str(" ")?;
                self.current_line += 1;
                line = &line[offset..];
            }
```
if there's no whitespace in `line[..self.cfg.line_len - self.current_line]` we uh, advance by zero, write a newline, then try again. i held down `z` for a while to see if the new 16550 code does anything weird on faster inputs and noticed uh, this.

on one hand, https://github.com/hawkw/mycelium/issues/493, on the other hand, the wrapping probably makes sense for non-serial-port readers (like, it's better to wrap smartly in the framebuffer than dumbly?)

Issue #493, opened 2024-12-29
get rid of serial-port line wrapping logic


hawkw:
sometimes it makes decisions that just look really terrible. for example:
```
[     ?.??????][?] │┌LocalApic::start_periodic_timer: interval=10ms
[     ?.??????][*] │├hal_x86_64::interrupt::apic::local: CPUID contains hypervisor info
[     ?.??????][*] │├hal_x86_64::interrupt::apic::local: hypervisor.apic_frequency=Some(
                   ││     1000000,
                   ││ )
[     ?.??????][?] │├hal_x86_64::interrupt::apic::local: determ
                   ││ ined APIC frequency from CPUID hypervisor info, frequency_hz=62
```
why did you have to write
```
determ
ined APIC frequency
```

We should just print the whole log on one line I guess.

PR #551, opened 2025-09-15
write_str loop should terminate

iximeow:
i've got a sneaking suspicion that in the serial context the ANSI control characters are taking up invisible space, which definitely makes #493 look way goofier than it should:
image

but at least with this the above doesn't send the kernel off into a spin? anyway this seems to do the needful for #550.


hawkw (commented):
whoop's

hawkw (approved):

PR #561, opened 2026-03-12
fix(maitake-sync): don't enable portable-atomic/critical-section

okhsunrog:
## Summary
- The `critical-section` feature in both `maitake-sync` and `maitake` was enabling `portable-atomic/critical-section`, which conflicts with `portable-atomic/unsafe-assume-single-core` (used by e.g. esp-hal for ESP32-C3)
- Per [portable-atomic's docs](https://docs.rs/portable-atomic/latest/portable_atomic/#optional-cfg), libraries should leave the portable-atomic backend choice to the end user
- Updated the no-atomics CI check to explicitly pass `portable-atomic/critical-section`, simulating what an end user on a no-CAS target would do

Fixes #559

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -284,7 +284,11 @@ jobs:
       # targets without native CAS atomics, portable-atomic needs a backend.
       # The maitake-sync crate intentionally does NOT enable this feature
       # itself (see hawkw/mycelium#559), leaving the choice to the end user.
-      run: cargo check -p maitake-sync --target thumbv6m-none-eabi --no-default-features --features=critical-section,portable-atomic/critical-section
+      run: >-
+        cargo check -p maitake-sync
+        --target thumbv6m-none-eabi
+        --no-default-features
+        --features=critical-section,portable-atomic/critical-section
 
   # test with `--no-default-features`
   maitake_no_default_features:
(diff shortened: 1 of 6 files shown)


hawkw (commented):
Overall, I think I'm okay with this change, as I discussed here:

However, there are a few additional notes before we can merge this. In particular, our README currently states that the `critical-section` feature flag _also enables `portable-atomic/critical-section`_, which it now no longer does:

https://github.com/hawkw/mycelium/blob/d13c2e561669b2e7b145149fc0f1f74ff2785e0f/maitake-sync/README.md?plain=1#L217

We should remove that --- and maybe add an explicit note that we *don't* enable it in the documentation for this feature.

We may also want to add some more discussion in the documentation here:

https://github.com/hawkw/mycelium/blob/d13c2e561669b2e7b145149fc0f1f74ff2785e0f/maitake-sync/README.md?plain=1#L118-L206

I just want to make sure that users understand what they have to do to get everything using either `critical-section` or whatever else is required on their platform consistently.

okhsunrog:
I think the CI needs an approval to run

hawkw (approved):
Thanks for the documentation changes, I think this looks good. I'll publish a release once this makes it through CI.

PR #565, opened 2026-07-29
feat(cordyceps): add `TransferStack::is_empty`

cedrivard:
Adds a non-mutating emptiness check to `TransferStack`:

```rust
#[inline]
#[must_use]
pub fn is_empty(&self) -> bool {
    self.head.load(Relaxed).is_null()
}
```

`TransferStack` already lives entirely under `#[cfg(target_has_atomic = "ptr")]`, so this
introduces no CAS/non-CAS asymmetry.

The result is documented as a snapshot: a producer may push the instant it returns, and a
`false` result must still be followed by `take_all`, whose `AcqRel` swap is what
synchronizes with the pushes it takes.

## Motivation

The only consumer-side operation on `TransferStack` today is `take_all`, a `swap` on the
head pointer. A consumer polling a usually-empty stack therefore performs a
read-modify-write on every check, including the checks that find nothing. `is_empty`
allows a plain load first.

The concrete case is `embassy-executor`, which uses `TransferStack` as its task run queue.
`RunQueue::dequeue_all` calls `take_all` unconditionally on every scheduler pass, so an
idle executor still issues an RMW and dirties that cache line on every wakeup. Skipping it
saves a write and avoids cache-line ping-pong between an idling core and a core enqueuing
work.

On RP2350 the effect is larger. That part has an erratum where any exclusive access
(`ldaex`/`stlex`) raises a global-monitor event acting as an effective `SEV`. The event
flag is sticky, so the `WFE` closing embassy's thread-executor idle loop always had an
event pending and returned immediately: the executor free-ran at roughly 1 MHz and never
slept. The unconditional `swap` was one of the exclusives keeping that loop dirty.
Measured on an RP2350 firmware after removing it:

|        | passes/s | polls/s | polls/pass | busy% |
|--------|---------:|--------:|-----------:|------:|
| before |   850993 |    1832 |       0.00 |  23.6 |
| after  |     3523 |    1923 |       0.55 |  12.7 |

Idle draw with only a heartbeat task running fell from 27.65 mA to 20.35 mA.

Refs: embassy-rs/embassy#4818, raspberrypi/pico-feedback#482, raspberrypi/pico-sdk#1812.

`embassy-executor` depends on `cordyceps = "0.3.4"`, so a new release would be what lets
the optimization land there. Would you consider cutting a v0.3.5 once this merges?


hawkw (approved):
Sure, this seems reasonable to me!

cedrivard:
Thanks for the quick review and merge, and for taking a change driven by someone else's use case.
Much appreciated :+1: 

My embassy-executor half needs another pass before it is fit to show anyone, so I will be watching for the next cordyceps release in the meantime.

hawkw:
> Thanks for the quick review and merge, and for taking a change driven by someone else's use case.
> Much appreciated 👍

Happy to, it was a pretty straightforward change and seemed like it was worth having! Here's [`cordyceps` v0.3.5](https://crates.io/crates/cordyceps/0.3.5); enjoy!

